PR #516, opened 2023-09-25
OSS validations

jmg-duarte:


jmg-duarte:
Temporarily closing just to stop running the validations non-stop

Makefile
```diff
@@ -456,8 +453,6 @@ rust/actyx/target/$(TARGET)/release/%: cargo-init make-always
 	  -w /src/rust/actyx \
 	  -e HOME=/home/builder \
 	  -v `pwd`:/src \
-	  -v $(CARGO_HOME)/for_builder/git:/home/builder/.cargo/git \
-	  -v $(CARGO_HOME)/for_builder/registry:/home/builder/.cargo/registry \
```
rkuhn, line 460: not caching these leads to significant download times, but I guess we don’t have persistence on the runners anyway?

jmg-duarte: We don't (at least for now) we could store using GitHub artifacts/storage but that should be investigated later

jmg-duarte:
I'm closing this until we need it again

PR #521, opened 2023-09-28
Fix dependabot alerts

jmg-duarte:


rkuhn (approved):

jmg-duarte:
I don't remember why I didn't merge this but now it "doesn't apply" so I'm closing it

PR #524, opened 2023-10-02
fix json path setting issue

Kelerchian:


js/node-manager/src/client/screens/Settings.tsx
```diff
@@ -1,14 +1,31 @@
...
+const setOrReplace = (s: object, path: string, val: unknown): object => {
+  // parse:
+  //  "/a/b" => ['a', 'b']
+  //  "/" => []
+  //  "" => []
+  const pointer = parse(path.trim()).filter((x) => !!x)
+  if (pointer.length === 0) {
+    const c: { default: object } = { default: s }
+    set(c, '/default', val)
+    return c.default
```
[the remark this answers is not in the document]
Kelerchian: > Could you explain what's going on "behind the scenes" here?

I don't want to get into the implementation details of https://www.npmjs.com/package/json-pointer, so I can't.

There may be subvalue merging inside. It might be just purely a replacement. 
But by using `set` instead of returning, I can be sure that it will behave the same as the `else` branch.

js/node-manager/src/client/screens/Settings.tsx
```diff
@@ -1,14 +1,31 @@
...
 import { validateAgainstSchema } from '../components/SettingsEditor'
 
+// Handles case where `set` cannot set root path
+// Because require('json-pointer').set throws when `path` is empty
+const setOrReplace = (s: object, path: string, val: unknown): object => {
+  // parse:
+  //  "/a/b" => ['a', 'b']
+  //  "/" => []
+  //  "" => []
+  const pointer = parse(path.trim()).filter((x) => !!x)
```
rkuhn, line 18: The `!!x` smells wrong to me: what if a path component is `'false'`?

Kelerchian: Per NodeJS 16

`!!false === false`
`!!'false' === true`

The issue where `'false'` string is regarded as falsy has been gone for I think years or decades.
(Even I am almost convinced that it is an instance of Mandela Effect because I can't find it anywhere online, but it might be there in PHP instead because a long time ago I am sure I had the same problem)

Do we use `'false'` string anywhere btw?

jmg-duarte (approved):

js/node-manager/src/client/screens/Settings.tsx
Kelerchian: Ah, I didn't realize that this indirection is confusing to some people.
I thought this implication was universal because '' is the only falsy string there is in JS.
Will change to that

rkuhn: well, `null` or `undefined` are also falsy, IIRC

Kelerchian: I mean, falsy string since `parse` returns string[].
Fixed it

rkuhn (approved):

PR #527, opened 2023-10-03
Fix fully_meshed on single core

Kelerchian:


rust/actyx/swarm/harness/src/lib.rs
```diff
@@ -316,7 +317,21 @@ where
...
-                        let f = machine.select(|ev| m!(ev.borrow(), Event::Connected(p) => *p));
+                        let f = async {
+                            let x = machine.select(|ev| m!(ev.borrow(), Event::Connected(p) => *p));
+                            futures::pin_mut!(x);
+                            loop {
+                                let poll = futures::poll!(&mut x);
+                                match poll {
+                                    task::Poll::Ready(x) => return x,
+                                    task::Poll::Pending => {
+                                        tokio::task::yield_now().await;
```
rkuhn, line 328: if this fixes it then the real fix is in `machine.select`

Kelerchian: This does not completely fix it unfortunately, at least with single core.

rust/actyx/swarm/harness/src/lib.rs
```diff
@@ -299,42 +300,74 @@ pub async fn fully_meshed<E>(sim: &mut Netsim<Command, E>, timeout: Duration) ->
...
+                                };
+                            }
+                        };
+
+                        futures::pin_mut!(f);
+                        match select(deadline.as_mut(), f).await {
+                            Either::Left(_) => Either::Left(()),
+                            Either::Right(r) => Either::Right(r),
+                        }
+                    };
```
rkuhn, line 340: This should really just be
```rust
let f = machine.select(...);
futures::pin_mut!(f);
let res = select(deadline.as_mut(), f).await;
```
If changing it to the code above changes the behaviour, then something really weird must be broken in strange ways. And the breakage is most likely in `machine.select`.

Kelerchian: This got me a compile error due to multiple mutable references, But I'll remove the inner polling future

rust/actyx/swarm/cli/src/main.rs
```diff
@@ -128,6 +152,16 @@ async fn run() -> Result<()> {
...
+                // NOTE: ipfs_embed::Event::Connected is not always emitted
+                // Therefore ipfs_embed::Event::ConnectionEstablished is used as a fallback
+                // See:
+                //  - https://docs.rs/crate/ipfs-embed/latest/source/src/net/peers.rs#:~:text=self.notify(Event%3A%3AConnected(c.peer_id))%3B
+                //      Connected event is SOMETIMES emitted, that is only when `event.other_is_established == 0`.
+                //  - https://docs.rs/crate/libp2p-swarm/0.41.1/source/src/lib.rs#:~:text=let%20non_banned_established
+                //      other_established_connection_ids - banned_peers
+                //  - https://docs.rs/crate/libp2p-swarm/0.41.1/source/src/lib.rs#:~:text=let%20non_banned_established
+                //      other_established_connection_ids is the
+                ipfs_embed::Event::ConnectionEstablished(peer_id, _) => Some(Event::Connected(peer_id)),
```
rkuhn, line 164: With the above workaround we shouldn’t need this line and the Connected line as well.

rkuhn (approved):
Great work narrowing it down to this detail! I think this is good enough for now, the proper fix needs to be done in ipfs-embed, but that can be done asynchronously.

rust/actyx/swarm/cli/src/main.rs
Kelerchian: Can we comment those lines instead of removing them for future context's sake?

rkuhn: sure!

rkuhn (approved):
LGTM (with the final nitpick)

rust/actyx/swarm/cli/src/main.rs
Kelerchian: comments added

PR #532, opened 2023-10-05
add generic MultiplexedWebsocket request tracker

Kelerchian:


js/sdk/src/v2/multiplexedWebsocket.ts
```diff
@@ -68,6 +75,13 @@ export class MultiplexedWebsocket {
           : throwError(() => err)
       }),
     )
+
+    activeRequestInternals.register(sym, {
+      serviceId: serviceId,
+      observable: request,
```
rkuhn, line 81: What’s the intended use of the response observable here? As a user I’d want to have the request timestamp and the payload in here.

rkuhn (commented):
Yes, apart from some textual linting looks like a good start.

js/sdk/src/v2/multiplexedWebsocket.ts
Kelerchian: I am not sure. Initially, I thought this was useful if the app wanted to close the request.
I'll add a timestamp and payload.

rkuhn: My design idea of this feature is only related to observability. Making it possible to influence the behaviour of the app via this route sounds dangerous to me: cancellation and timeouts should be done by the code performing the requests, not via a global entrypoint.

js/sdk/src/globals.ts
```diff
@@ -0,0 +1,41 @@
...
+type ActiveRequestGlobals = {
+  all: () => ActiveRequest[]
+  [GlobalInternalSymbol]: {
+    register: (sym: symbol, activeRequest: ActiveRequest) => unknown
+    unregister: (sym: symbol) => unknown
+  }
+}
+
+/**
+ * Global object for accessing currently active requests
```
rkuhn, line 18: Regarding globals: it would make sense from an API standpoint to offer the `activeRequests` from each SDK instance. A global variable is only needed to alert the user when they creates multiple SDK instances, which is almost always a mistake.

rkuhn (commented):
LGTM apart from the question of global vs. per-SDK

js/sdk/src/globals.ts
Kelerchian: My initial thought is that with `globals` the user will be able to write it in the global/static scope.
I often see this pattern for telemetry purposes, which is written before the main function.

As with multiple SDKs, I think we should go with `console.warn`-ing instead when 2 or more instances are created.

rkuhn: okay, makes sense

PR #533, opened 2023-10-05
Add --expires-in to ax-certs

### LINKED ISSUE #512: Handle human-friendly dates on `--expires-at` for ax-cert

jmg-duarte:
closes #512 

I'm showing both commits so you can pick between them, I didn't benchmark it but this is for a "single-use command" so I'm more interested in what you think is more readable

18f5d6120795ea801b29bf1f85072d34440bac2b uses a regex for parsing 
d26525e8e2a394d493e725c7cadca23e25f37690 uses _Just Code TM_ for parsing (this commit is actually buggy but after fixing the errors it passes the tests as well)

### DIFF
--- a/rust/actyx/certs/src/bin/ax-cert.rs
+++ b/rust/actyx/certs/src/bin/ax-cert.rs
@@ -1,11 +1,9 @@
-use std::str::FromStr;
-
 use actyx_sdk::AppId;
 use certs::{AppDomain, DeveloperCertificate, SignedAppLicense};
 use chrono::{DateTime, Utc};
 use crypto::PrivateKey;
 use lazy_static::lazy_static;
-use regex::{Match, Regex};
+use regex::Regex;
 use structopt::StructOpt;
 use util::version::NodeVersion;
 
@@ -57,8 +55,13 @@ struct AppLicenseOpts {
 /// m - minute(s)
 /// s - second(s)
 fn parse_expires_in(expires_in: &str) -> Result<DateTime<Utc>, anyhow::Error> {
+    parse_expires_in_as_duration(expires_in).map(|d| DateTime::from(Utc::now() + d))
+}
+
+// This function is much easier to test for correctness
+fn parse_expires_in_as_duration(expires_in: &str) -> Result<chrono::Duration, anyhow::Error> {
     lazy_static! {
-        static ref RE: Regex = Regex::new(r"((?P<years>[0-9]+)Y)?\s*((?P<months>[0-9]+)M)?\s*((?P<weeks>[0-9]+)w)?\s*((?P<days>[0-9]+)d)?\s*((?P<hours>[0-9]+)h)?\s*((?P<minutes>[0-9]+)m)?\s*((?P<seconds>[0-9]+)s)?\s*").unwrap();
+        static ref RE: Regex = Regex::new(r"\s*((?P<years>[0-9]+)Y)?\s*((?P<months>[0-9]+)M)?\s*((?P<weeks>[0-9]+)w)?\s*((?P<days>[0-9]+)d)?\s*((?P<hours>[0-9]+)h)?\s*((?P<minutes>[0-9]+)m)?\s*((?P<seconds>[0-9]+)s)?\s*").unwrap();
     }
 
     let captures = RE
@@ -75,42 +78,42 @@ fn parse_expires_in(expires_in: &str) -> Result<DateTime<Utc>, anyhow::Error> {
         );
     duration = duration
         + chrono::Duration::days(
-            365 * captures
+            30 * captures
                 .name("months")
                 .and_then(|m| m.as_str().parse::<i64>().ok())
                 .unwrap_or(0),
         );
     duration = duration
         + chrono::Duration::days(
-            365 * captures
+            7 * captures
                 .name("weeks")
                 .and_then(|m| m.as_str().parse::<i64>().ok())
                 .unwrap_or(0),
         );
     duration = duration
         + chrono::Duration::days(
-            365 * captures
+            captures
                 .name("days")
                 .and_then(|m| m.as_str().parse::<i64>().ok())
                 .unwrap_or(0),
         );
     duration = duration
-        + chrono::Duration::days(
-            365 * captures
+        + chrono::Duration::hours(
+            captures
                 .name("hours")
                 .and_then(|m| m.as_str().parse::<i64>().ok())
                 .unwrap_or(0),
         );
     duration = duration
-        + chrono::Duration::days(
-            365 * captures
+        + chrono::Duration::minutes(
+            captures
                 .name("minutes")
                 .and_then(|m| m.as_str().parse::<i64>().ok())
                 .unwrap_or(0),
         );
     duration = duration
-        + chrono::Duration::days(
-            365 * captures
+        + chrono::Duration::seconds(
+            captures
                 .name("seconds")
                 .and_then(|m| m.as_str().parse::<i64>().ok())
                 .unwrap_or(0),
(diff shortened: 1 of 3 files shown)


jmg-duarte:
@Kelerchian that's `--expires-at`, this is for friendlier UX

Kelerchian:
Ah, my bad. I saw the PR title and just the major bulk of the changes.
"Just Code" is easier to see IMO

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -32,13 +34,97 @@ struct AppLicenseOpts {
...
+    #[structopt(long, parse(try_from_str = parse_expires_in))]
+    expires_in: Option<DateTime<Utc>>,
 
     /// Requester's email address
     #[structopt(long)]
     email: String,
 }
 
+/// Parsing function for the `expires_in` variable.
+/// Accepts strings matching the regex: ^([0-9]+(Y|M|w|d|h|m|s))+$
```
rkuhn, line 48: this isn’t true: it only finds one of each, and only in descending order or suffix magnitude

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -32,13 +34,97 @@ struct AppLicenseOpts {
 
     /// ISO 8601 (i.e. 2014-11-28T12:00:09Z) expiration date time
     #[structopt(long)]
-    expires_at: DateTime<Utc>,
+    expires_at: Option<DateTime<Utc>>,
+
+    #[structopt(long, parse(try_from_str = parse_expires_in))]
```
rkuhn, line 39: please add a doc comment above (will be displayed in `-h` / `--help` output)

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -32,13 +34,97 @@ struct AppLicenseOpts {
...
+        .captures(expires_in)
+        .ok_or_else(|| anyhow::anyhow!("Failed to parse string."))?;
+
+    let mut duration = chrono::Duration::zero();
+    duration = duration
+        + chrono::Duration::days(
+            365 * captures
+                .name("years")
+                .and_then(|m| m.as_str().parse::<i64>().ok())
+                .unwrap_or(0),
```
rkuhn, line 77: how about wrapping this in a local `fn get(c: &Captures, n: &str) -> i64`?

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -32,13 +34,97 @@ struct AppLicenseOpts {
...
+    duration = duration
+        + chrono::Duration::seconds(
+            captures
+                .name("seconds")
+                .and_then(|m| m.as_str().parse::<i64>().ok())
+                .unwrap_or(0),
+        );
+
+    if duration.is_zero() {
+        return Err(anyhow::anyhow!("Expiration time must be bigger than zero"));
```
rkuhn, line 123:
```suggestion
        return Err(anyhow::anyhow!("Expiration interval must be bigger than zero"));
```

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -107,9 +193,111 @@ fn create_dev_cert(opts: DevCertOpts) -> anyhow::Result<()> {
 }
 
 fn create_app_license(opts: AppLicenseOpts) -> anyhow::Result<()> {
-    let license = SignedAppLicense::new(opts.actyx_private_key, opts.email, opts.app_id, opts.expires_at, None)?;
+    let expiration_date = opts
+        .expires_at
+        .or(opts.expires_in)
+        .unwrap_or_else(|| Utc::now() + chrono::Duration::days(1));
```
rkuhn, line 199: either document the default of one day or remove it and raise an error if both args are absent

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -107,9 +193,111 @@ fn create_dev_cert(opts: DevCertOpts) -> anyhow::Result<()> {
...
     println!("{}", serialized);
 
     Ok(())
 }
+
+#[cfg(test)]
+mod test_expires_in {
+    use crate::parse_expires_in_as_duration;
+
+    // NOTE(duarte): Quickcheck would probably be amazing to test this but I don't have the time to learn it now
```
rkuhn, line 212: perhaps instead add a smoke test that uses a sequence like `2Y1w15h`

rkuhn (commented):
I only reviewed the regex version, that’s a reasonable approach

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -32,13 +34,97 @@ struct AppLicenseOpts {
...
+/// m - minute(s)
+/// s - second(s)
+fn parse_expires_in(expires_in: &str) -> Result<DateTime<Utc>, anyhow::Error> {
+    parse_expires_in_as_duration(expires_in).map(|d| Utc::now() + d)
+}
+
+// This function is much easier to test for correctness
+fn parse_expires_in_as_duration(expires_in: &str) -> Result<chrono::Duration, anyhow::Error> {
+    lazy_static! {
+        static ref RE: Regex = Regex::new(r"\s*((?P<years>[0-9]+)Y)?\s*((?P<months>[0-9]+)M)?\s*((?P<weeks>[0-9]+)w)?\s*((?P<days>[0-9]+)d)?\s*((?P<hours>[0-9]+)h)?\s*((?P<minutes>[0-9]+)m)?\s*((?P<seconds>[0-9]+)s)?\s*").unwrap();
```
rkuhn, line 64: perhaps split across multiple lines:
```suggestion
        static ref RE: Regex = Regex::new(r"\s*\
          ((?P<years>[0-9]+)Y)?\s*\
          ((?P<months>[0-9]+)M)?\s*\
          ((?P<weeks>[0-9]+)w)?\s*\
          ((?P<days>[0-9]+)d)?\s*\
          ((?P<hours>[0-9]+)h)?\s*\
          ((?P<minutes>[0-9]+)m)?\s*\
          ((?P<seconds>[0-9]+)s)?\s*").unwrap();
```

rust/actyx/certs/src/bin/ax-cert.rs
jmg-duarte: The comment rot, nice catch

rust/actyx/certs/src/bin/ax-cert.rs
jmg-duarte: Would've been amazing if it worked. With raw string literals it doesn't. But I found another solution

rust/actyx/certs/src/bin/ax-cert.rs
jmg-duarte: Documented the precedence of `--expires-at` and removed the default. Thanks for the suggestion!

rust/actyx/certs/src/bin/ax-cert.rs
```diff
@@ -30,15 +32,90 @@ struct AppLicenseOpts {
...
+/// h - hour(s)
+/// m - minute(s)
+/// s - second(s)
+fn parse_expires_in(expires_in: &str) -> Result<DateTime<Utc>, anyhow::Error> {
+    parse_expires_in_as_duration(expires_in).map(|d| Utc::now() + d)
+}
+
+// This function is much easier to test for correctness
+fn parse_expires_in_as_duration(expires_in: &str) -> Result<chrono::Duration, anyhow::Error> {
+    lazy_static! {
```
Kelerchian, line 76: curious, why does this need to be a lazy_static?

rkuhn: Because only trivial (i.e. `const`) initializers are allowed for `static` names, so the regex needs to be parsed and compiled upon first use.

rkuhn (approved):

rust/actyx/certs/src/bin/ax-cert.rs
Kelerchian: ah i see. thanks for the explanation

jmg-duarte: (I think, lmk if this is wrong) In this case the performance difference is irrelevant, we're matching the value once, so using `lazy_static` or just the `Regex::new` will be pretty much the same. I did it more out of an habit rather than "this is performance critical and I need this here"

Kelerchian: Understandable and I don't see why not.
Since, if the function is moved and reused eventually this will be a boon.

PR #539, opened 2023-10-06
Fix MacOS build

jmg-duarte:
I didn't push the MacOS image because we're moving off of self-hosted, but while we don't, we still need the builds to work


jmg-duarte:
Fixed build BTW https://github.com/Actyx/Actyx/actions/runs/6431800981

rkuhn (approved):
the difference between the two images is the user ID, right?

jmg-duarte:
> the difference between the two images is the user ID, right?

Yes

PR #545, opened 2023-10-11
Rollback the publishing method

jmg-duarte:
The manifest create & push mechanism is a way of
"promoting images to production", by creating a manifest and pushing it under the
the name of another image, Docker will essentially just re-use the underlying images
to create a new one, under a new namespace. This is *crucial* for us because we
check if Docker releases are missing using the manifests.

### DIFF
--- a/rust/release/src/publisher.rs
+++ b/rust/release/src/publisher.rs
@@ -94,18 +94,20 @@ impl Publisher {
     }
     pub fn source_exists(&self) -> anyhow::Result<bool> {
         log::debug!("checking if source {} exists for target {}", self.source, self.target);
-        match &self.source.r#type {
+        let result = match &self.source.r#type {
             SourceType::Blob(s) => blob_exists(Container::Artifacts, s),
             SourceType::Docker {
                 registry,
                 tag,
                 repository,
             } => docker_manifest_exists(&format!("{}/{}:{}", registry, repository, tag), None),
-        }
+        };
+        log::debug!("source checking result: {:?}", result);
+        result
     }
     pub fn target_exists(&self) -> anyhow::Result<bool> {
         log::debug!("checking if target {} exists for source {}", self.target, self.source);
-        match &self.target {
+        let result = match &self.target {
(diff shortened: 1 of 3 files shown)


rust/release/src/publisher.rs
```diff
@@ -171,46 +175,21 @@ impl Publisher {
...
-                    anyhow::ensure!(
-                        cmd.status.success(),
-                        "Error running `docker {:?}` for {}\nstdout: {}\nstderr: {}",
-                        args,
-                        target,
-                        String::from_utf8(cmd.stdout)?,
-                        String::from_utf8(cmd.stderr)?
-                    );
-                    Ok(())
+                    docker_manifest_create(manifest, registry, source_repository, repository, tag)
```
Kelerchian, line 185: This function seems to call `manifest create`, replacing what previously calls `buildx build`.

When will `buildx build` be called?

rust/release/src/publisher.rs
```diff
@@ -748,6 +699,68 @@ fn docker_manifest_rm(target: &str) -> anyhow::Result<()> {
...
+    target_repository: &str,
+    tag: &str,
+) -> anyhow::Result<()> {
+    let target = format!("{}/{}:{}", registry, target_repository, tag);
+    // Make sure the manifest is removed locally, otherwise we can't
+    // (re)create it
+    docker_manifest_rm(&target)?;
+
+    // Now the actual manifest creation
+    let args = vec!["manifest".to_string(), "create".to_string(), target.clone()]
```
rkuhn, line 738: I’d leave off the `vec!` as an array should work as well, right?

rkuhn (approved):
LGTM

rust/release/src/publisher.rs
jmg-duarte: Not in place (otherwise Clippy would've complained). 
From my understanding:
- The array's `into_iter` is just a call to `iter`
- Meaning it doesn't take ownership of the contents of the iterator 
- Thus it needs to be chained with `&String`s instead of `String`s
- These are dropped because they're local references 
Meaning it's a PITA to avoid one `vec!`

PR #546, opened 2023-10-12
Brief housekeeping in the SDK

jmg-duarte:
- Rename the `HttpClient` to `ActyxClient`
- Remove the thread-based exponential backoff

### DIFF
--- a/rust/sdk/src/arb.rs
+++ b/rust/sdk/src/arb.rs
@@ -1,3 +1,7 @@
+// Used by trees::arb
+// Can't really be moved to `trees` since Rust disallows
+// implementing foreign traits on foreign types
+
 use std::collections::BTreeMap;
 
 use crate::{
--- a/rust/sdk/src/client.rs
+++ b/rust/sdk/src/client.rs
@@ -261,6 +261,12 @@ impl Debug for ActyxClient {
 
 #[async_trait]
 impl EventService for ActyxClient {
+    /// Returns known offsets across local and replicated streams.
+    ///
+    /// If an authorization error (code 401) is returned, it will try to re-authenticate.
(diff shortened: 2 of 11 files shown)


rust/sdk/src/arb.rs
```diff
@@ -1,3 +1,7 @@
+// Used by trees::arb
+// Can't really be moved to `trees` since Rust disallows
+// implementing foreign traits on foreign types
```
rkuhn, line 3: yeah, this is a peculiarity of the Rust way of doing things, which leads to good libraries implementing Arbitrary even though they don’t use it internally

rust/sdk/src/client.rs
```diff
@@ -115,8 +115,11 @@ impl HttpClient {
         Ok(token)
     }
 
-    /// Makes request to Actyx apis. On http authorization error tries to
-    /// re-authenticate and retries the request ten times with increasing delay.
+    /// Perform a request (to Actyx APIs).
+    /// If an authorization error (code 401) is returned, it will try to re-authenticate.
+    /// If the service is unavailable (code 503), this method will retry to perform the
+    /// request up to 10 times with exponentially increasing delay - currently,
+    /// this behavior is only available if the `with-tokio` feature is enabled.
```
rkuhn, line 122: This text is good, but on a private method people won’t see it. I’d probably add it to each end-user method that calls this function.

rkuhn (approved):
LGTM apart from the doc comment

rust/sdk/src/client.rs
jmg-duarte: I can add on them as well, this is a rust doc and not a regular one because we're the first clients after all

PR #547, opened 2023-10-12
Remove JS SDK v1

jmg-duarte:
I was expecting to remove the `SnapshotStore` but it's still used in `EventFnsFromEventStoreV2`


js/sdk/src/actyx.ts
```diff
@@ -124,10 +108,7 @@ export const Actyx = {
     log.actyx.debug('NodeId call returned:', nodeId)
 
     if (!nodeId) {
-      // Try connecting to v1 if we failed to retrieve a v2 node id
-      // (Note that if the port is completely unreachable, v2getNodeId will throw an exception and we don’t get here.)
-      log.actyx.debug('NodeId was null, trying to reach V1 backend...')
-      return createV1(opts)
+      throw new Error('Node ID is missing')
```
rkuhn, line 111: I assume this is reached when there was a HTTP server but it returned the wrong response — it would help debugging such incidents to mention this in the error message.

Kelerchian:
Not sure if this should be done now or later: bumping the version in `package.json` (and `npm install`-ing) afterwards before creating a commit to update `package-lock.json`

js/sdk/src/actyx.ts
jmg-duarte: I can improve the call to `v2getNodeId`:
```ts
export const v2getNodeId = async (config: ActyxOpts): Promise<string | null> => {
  const path = `http://${getApiLocation(config.actyxHost, config.actyxPort)}/node/id`
  return await fetch(path)
    .then((resp) => {
      // null indicates the endpoint was reachable but did not react with OK response -> probably V1.
      return resp.ok ? resp.text() : null
    })
    .catch((err) => {
      // ECONNREFUSED is probably not a CORS issue, at least...
      if (err.message && err.message.includes('ECONNREFUSED')) {
        throw new Error(decorateEConnRefused(err.message, path))
      }

      log.actyx.info(
        'Attempt to connect to V2 failed with unclear cause. Gonna try go connect to V1 now. Error was:',
        err,
      )
      // HACK: V1 has broken CORS policy, this blocks our request if it reaches the WS port (4243) instead of the default port (4454).
      // So if we got an error, but the error is (probably) not due to the port being closed, we assume: Probably V1.
      // (Would be awesome if JS API gave a clear and proper indication of CORS block...)
      return null
    })
}
```
Maybe instead of checking for connection refused, we just remove the `catch` block so the error propagates up?

PR #551, opened 2023-10-13
fold ax-cert into ax plus improvements

rkuhn:
feat(cli): add generation of licenses and developer certificates
feat(cli): use AX_IDENTITY environment variable when -i is not given
feat(cli): allow passing of private key instead of identity filename

### DIFF
--- a/rust/actyx/ax/src/cmd/apps/license.rs
+++ b/rust/actyx/ax/src/cmd/apps/license.rs
@@ -1,43 +1,35 @@
+use crate::cmd::AxCliCommand;
 use actyx_sdk::AppId;
-use certs::{AppDomain, DeveloperCertificate, SignedAppLicense};
+use certs::SignedAppLicense;
 use chrono::{DateTime, Utc};
 use crypto::PrivateKey;
+use futures::{stream::once, FutureExt, Stream};
 use lazy_static::lazy_static;
-use regex::{Captures, Regex};
+use regex::Regex;
 use structopt::StructOpt;
-use util::version::NodeVersion;
+use util::formats::{ActyxOSCode, ActyxOSError, ActyxOSResult, ActyxOSResultExt};
 
 #[derive(StructOpt, Debug)]
-struct DevCertOpts {
-    #[structopt(long, env, hide_env_values = true)]
-    /// Actyx private key
-    actyx_private_key: PrivateKey,
+#[structopt(version = env!("AX_CLI_VERSION"))]
+pub struct LicenseOpts {
+    /// The secret key used to sign the license
+    /// (this must match the AX_PUBLIC_KEY your `actyx` binary has been compiled with).
+    #[structopt(long, short = "A", env, hide_env_values = true)]
+    ax_secret_key: PrivateKey,
 
-    #[structopt(long, env, hide_env_values = true)]
-    /// Developer's private key, if omitted one will be generated
-    dev_private_key: Option<PrivateKey>,
-
-    /// Certificate's allowed app domains
-    #[structopt(long, required = true)]
-    app_domains: Vec<String>,
-}
-
-#[derive(StructOpt, Debug)]
-struct AppLicenseOpts {
-    #[structopt(long, env, hide_env_values = true)]
-    /// Actyx private key
-    actyx_private_key: PrivateKey,
-
-    /// App id
+    /// The app id of the app to create a license for,
+    /// use `com.actyx.node` to create a node license.
     #[structopt(long)]
     app_id: AppId,
 
-    /// An expiration date time in ISO 8601 (i.e. 2014-11-28T12:00:09Z).
-    /// Takes precedences over `--expires-in`.
+    /// An expiration date time in ISO 8601 (i.e. 2014-11-28T12:00:09Z),
+    /// takes precedence over `--expires-in`.
     #[structopt(long)]
     expires_at: Option<DateTime<Utc>>,
 
-    /// The amount of time in which the license should expire. The accepted format is
+    /// The amount of time in which the license should expire.
+    ///
+    /// The accepted format is
     /// composed of a number followed by a unit (i.e. "10Y"), whitespace is supported
     /// before, between and after the number and the unit. Values are accepted in
     /// descending order according to the size of the unit, the unit character and
@@ -48,12 +40,39 @@ struct AppLicenseOpts {
     /// Note: Years are considered to be 365 days long and months to be 30 days long.
     ///
     /// For example: "1Y 3M 4h" means that the license should expire in 1 year, 3 months and 4 hours.
-    #[structopt(long, parse(try_from_str = parse_expires_in))]
+    #[structopt(long, short = "e", parse(try_from_str = parse_expires_in))]
     expires_in: Option<DateTime<Utc>>,
 
     /// Requester's email address
     #[structopt(long)]
     email: String,
+}
+
+pub struct AppsLicense;
+
+impl AxCliCommand for AppsLicense {
+    type Opt = LicenseOpts;
+    type Output = String;
+
+    fn run(opts: Self::Opt) -> Box<dyn Stream<Item = ActyxOSResult<Self::Output>> + Unpin> {
+        Box::new(once(
+            async move {
+                let expiration_date = opts.expires_at.or(opts.expires_in).ok_or(ActyxOSError::new(
+                    ActyxOSCode::ERR_INVALID_INPUT,
+                    "An expiration date must be specified. Use `--expires-at` or `--expires-in`.",
+                ))?;
+
+                let license = SignedAppLicense::new(opts.ax_secret_key, opts.email, opts.app_id, expiration_date, None)
+                    .ax_err(ActyxOSCode::ERR_INTERNAL_ERROR)?;
+                license.to_base64().ax_err(ActyxOSCode::ERR_INTERNAL_ERROR)
+            }
+            .boxed(),
+        ))
+    }
+
+    fn pretty(result: Self::Output) -> String {
+        result
+    }
 }
 
 /// Parsing function for the `expires_in` variable.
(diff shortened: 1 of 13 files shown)


rust/actyx/certs/src/lib.rs
jmg-duarte, line 1: I think we can do away with this extra serialization step, unless we're looking to provide an _exact_ reason as to why the deserialization of a signed manifest failed.

The trial serialization issue can be solved by removing code and moving the validation to [`validate_manifest`](https://github.com/Actyx/Actyx/blob/e4c1b33f4eea203c9387a088ec73cc2b1e62c943/rust/actyx/api/src/auth/mod.rs#L91-L101).
As demonstrated in https://github.com/Actyx/Actyx/commit/8a998ca0d8a2f2240f4125bf6c163885193abed6

rust/actyx/ax/src/cmd/apps/mod.rs
```diff
@@ -1,15 +1,20 @@
...
 use futures::Future;
 use structopt::StructOpt;
 
+use license::LicenseOpts;
 pub use sign::{create_signed_app_manifest, SignOpts};
 
 #[derive(StructOpt, Debug)]
 #[structopt(version = env!("AX_CLI_VERSION"))]
 /// manage app manifests
 pub enum AppsOpts {
```
jmg-duarte, line 14: I get the name was here before but since this manages app manifests and not apps themselves, I think it makes more sense to have this subcommand be renamed to `manifests`.

Thoughts?

rust/actyx/ax/src/cmd/apps/license.rs
```diff
@@ -91,113 +110,31 @@ fn parse_expires_in_as_duration(expires_in: &str) -> Result<chrono::Duration, an
...
-    duration = duration + chrono::Duration::hours(get_i64_from_capture(&captures, "hours"));
-    duration = duration + chrono::Duration::minutes(get_i64_from_capture(&captures, "minutes"));
-    duration = duration + chrono::Duration::seconds(get_i64_from_capture(&captures, "seconds"));
+    duration = duration + chrono::Duration::days(365 * get_i64_from_capture("years"));
+    duration = duration + chrono::Duration::days(30 * get_i64_from_capture("months"));
+    duration = duration + chrono::Duration::days(7 * get_i64_from_capture("weeks"));
+    duration = duration + chrono::Duration::days(get_i64_from_capture("days"));
+    duration = duration + chrono::Duration::hours(get_i64_from_capture("hours"));
+    duration = duration + chrono::Duration::minutes(get_i64_from_capture("minutes"));
+    duration = duration + chrono::Duration::seconds(get_i64_from_capture("seconds"));
```
jmg-duarte, line 127: I'm not a fan of this approach because it obscures where the captures are coming from. 
It requires one extra reading step to see that the `captures` are captured by the closure, instead of being very obvious where they come from.

rust/actyx/ax/src/cmd/users/pubkey.rs
```diff
@@ -0,0 +1,33 @@
...
+use crypto::PublicKey;
+use futures::{stream::once, FutureExt, Stream};
+use structopt::StructOpt;
+use util::formats::ActyxOSResult;
+
+#[derive(StructOpt, Debug)]
+#[structopt(version = env!("AX_CLI_VERSION"))]
+pub struct PubkeyOpts {
+    #[structopt(short, long, value_name = "FILE_OR_KEY", env = "AX_IDENTITY", hide_env_values = true)]
+    /// File from which the identity (private key) for authentication is read.
```
jmg-duarte, line 14: Question: Why add the comment after the attribute and not before?

jmg-duarte (commented):
Some comments, but nothing that blocks merging

rust/actyx/ax/src/cmd/users/pubkey.rs
rkuhn: It makes no difference, we can settle on the other order if you want (currently it isn’t consistent).

rust/actyx/certs/src/lib.rs
rkuhn: > … unless we're looking to provide an exact reason as to why the deserialization of a signed manifest failed.

This is the real reason from my point of view: with an untagged union the error message will always be unhelpful. Instead of the custom deserializer we could also just switch to a single manifest structure with optional signature.

rust/actyx/ax/src/cmd/apps/license.rs
rkuhn: My view is inspired by the same goal (readability) but diametrically different: I want to see as few redundant glyphs as possible so that the human eye concentrates on the essence of the differences between the lines. I’d prefer something like:

    duration += chrono::days(365 * get("years"));

rust/actyx/certs/src/lib.rs
jmg-duarte: > This is the real reason from my point of view: with an untagged union the error message will always be unhelpful.
True but unless we can provide better diagnostics for the signature too (which currently errors with a very similar message to "failed to deserialize") I find it overkill

> Instead of the custom deserializer we could also just switch to a single manifest structure with optional signature.
I like this approach, but would the manifest validation then return either a Trial or Signed?

rust/actyx/ax/src/cmd/users/pubkey.rs
jmg-duarte: Personally I prefer the comment above since it "improves reading locality":
```
comment
code (attribute)
code (actual code)
```

rust/actyx/ax/src/cmd/apps/license.rs
jmg-duarte: Agreed on the `+=` but that's a Chrono issue (see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3c541ba510954293916695f4397848e7)
I don't have a strong opinion on `get` vs  `get_i64_from_capture`. In these small cases it's always hard to vouch for one over the other without more people providing an opinion for better sampling :sweat_smile:

rust/actyx/ax/src/cmd/apps/mod.rs
rkuhn: This reasoning doesn’t apply to the `apps license` subcommand, though. And the intent of the `apps sign` subcommand is to indeed certify the app — I know that this is technically not possible right now, but it remains the intention.

rust/actyx/ax/src/cmd/apps/license.rs
rkuhn: Since `chrono::Duration` is indeed lacking I made an even more tailored utility function. I think those are fine if they easily fit on the same screen of their strictly scoped usage — which is also why I keep them inside the outer function instead of lifting them to the module level.

rust/actyx/ax/src/cmd/users/pubkey.rs
rkuhn: Okay, fixing it throughout.

rust/actyx/certs/src/lib.rs
rkuhn: Okay, since I don’t think optimising this representation has much benefit I’ll try to get rid of the enum completely.

rkuhn (commented):
@jmg-duarte I added a commit to unify the AppManifest without variants, it works nicely.

rust/actyx/api/src/auth/mod.rs
```diff
@@ -93,10 +93,11 @@ fn validate_manifest(
...
 ) -> Result<(AppMode, AppId, String), ApiError> {
-    match manifest {
-        AppManifest::Signed(x) => validate_signed_manifest(x, ax_public_key, licensing)
-            .map(|_| (AppMode::Signed, x.get_app_id(), x.version.clone())),
-        AppManifest::Trial(x) => Ok((AppMode::Trial, x.get_app_id(), x.version.clone())),
+    if manifest.is_signed() {
+        validate_signed_manifest(manifest, ax_public_key, licensing)
+            .map(|_| (AppMode::Signed, manifest.app_id(), manifest.version().to_owned()))
+    } else {
+        Ok((AppMode::Trial, manifest.app_id(), manifest.version().to_owned()))
```
jmg-duarte, line 100: I know the PR disallows creating "invalid" trials in the Rust code, but since this is part of a REST endpoint, we might be receiving requests from other SDKs, etc

Something among the lines of:
```suggestion
    } else if manifest.app_id().starts_with("com.example.") {
        Ok((AppMode::Trial, manifest.app_id(), manifest.version().to_owned()))
    } else {
        Err(ApiError::InvalidManifest { msg: "Trial application IDs are required to start with "com.example." }
    }
```

jmg-duarte (commented):
I'm confused as to why we need the wrapper type since they're usually used when we're trying to implement foreign traits for foreign types, while here we control `AppManifest`.

I left a comment on the `validate_manifest` because it seems to me it doesn't take into account that unsigned manifests can come from other SDKs.

rust/actyx/api/src/auth/mod.rs
rkuhn: This is exactly the reason for the wrapper type: AppManifestIo is the serialisation format that comes off the wire, which validates only the basic data types, while AppManifest firmly establishes the trial manifest validation rule (see `impl TryFrom`). I searched the serde github repo for confirmation, this is the intended implementation pattern of deserialisation-is-validation (i.e. `#[serde(validate = ...)]` were all rejected with this reason).

jmg-duarte (approved):

Kelerchian:
The constant errors mentioned by Jose [here](https://discord.com/channels/697349510409617408/1074986110024945705/1164581145518813314) comes from warp filter somehow err-ing
```
warp::filters::body[0m[2m:[0m request json body error: invalid type: null, expected a string at line 1 column 72
```
despite, the correct json being passed on the caller's side.  Will investigate more

rust/actyx/certs/src/app_manifest.rs
```diff
@@ -30,31 +30,65 @@ impl AppManifestSignature {
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let data = base64::decode(s).map_err(|_| "failed to base64 decode app manifest signature")?;
+        serde_cbor::from_slice::<AppManifestSignature>(&data)
+            .map_err(|_| "failed to deserialize to app manifest signature")
+    }
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+fn deserialize_signature<'de, D: Deserializer<'de>>(deserializer: D) -> Result<Option<AppManifestSignature>, D::Error> {
+    let s = String::deserialize(deserializer)?;
```
Kelerchian, line 49: Will this cause an error if signature field does not exist (e.g. because if `skip_serializing_if` flag below).
I figured that the failing harness test was due to failing AppManifest deserialization in `src/auth/mod.rs, route function`

Kelerchian: NVM, I just realized there are 2 different `AppManifest` s

PR #554, opened 2023-10-16
fix double-input issue in windows and add identity changing mechanism

Kelerchian:


rust/sdk/examples/chat/cmdline.rs
```diff
@@ -12,40 +12,115 @@ pub enum Cmdline {
...
-                    } else {
-                        text_area.write().input(key);
+                // Prevent double input because of KeyEventKind::Release
+                let key = match event {
+                    Event::Key(key) if key.kind == KeyEventKind::Press || key.kind == KeyEventKind::Repeat => key,
+                    _ => continue,
+                };
+
+                let is_editing_identity = { (identity.read()).edit.clone() };
+                match is_editing_identity {
```
rkuhn, line 71: I’d replace the match with

```rust
let edit_name = identity.read().edit.is_some();
if edit_name { ...
```

rust/sdk/examples/chat/cmdline.rs
```diff
@@ -12,40 +12,115 @@ pub enum Cmdline {
...
+                    Some(identity_buffer_value) => {
+                        {
+                            let mut identity = identity.write();
+                            match key.code {
+                                KeyCode::Esc => {
+                                    identity.edit = None;
+                                }
+                                KeyCode::Enter => {
+                                    identity.val = identity_buffer_value;
+                                    identity.edit = None;
```
rkuhn, line 82:
```suggestion
                                    identity.val = identity.edit.take();
```

rust/sdk/examples/chat/display.rs
```diff
@@ -127,6 +83,86 @@ pub async fn display(mut cell: ActoCell<Display, impl ActoRuntime>) {
...
+fn render<W: io::Write>(
+    f: &mut Frame<CrosstermBackend<W>>,
+    identity: &Reader<Identity>,
+    text_area: &Reader<TextArea>,
+    messages: &Reader<Vec<Message>>,
+    not_connected: &Option<String>,
+) {
+    identity.project(|identity| match &identity.edit {
+        Some(edit) => render_editing_identity(f, edit),
+        None => render_chat(f, text_area, messages, not_connected),
```
rkuhn, line 95: Remember that `identity.project()` holds a lock. Calling other functions that may obtain locks is best avoided, otherwise you get to analyse call trees and prove that no deadlocks can occur :-)

This can be fixed by returning `bool` from the projection function to indicate whether `render_chat` should still be called.

rust/sdk/examples/chat/display.rs
```diff
@@ -127,6 +83,86 @@ pub async fn display(mut cell: ActoCell<Display, impl ActoRuntime>) {
...
+    });
+}
+
+fn render_editing_identity<W: io::Write>(f: &mut Frame<CrosstermBackend<W>>, edit: &TextArea) {
+    let size = f.size();
+    let layout = Layout::default()
+        .direction(ratatui::prelude::Direction::Vertical)
+        .constraints([ratatui::prelude::Constraint::Length(1)])
```
rkuhn, line 103: does that put the edit block at the top of the terminal? should probably be in the middle

rkuhn (commented):
Nice improvements!

rust/sdk/examples/chat/display.rs
Kelerchian: It does. I don't know, I think text in the middle of a terminal is unnatural. But you can convince me otherwise

rust/sdk/examples/chat/display.rs
Kelerchian: Ah yes. For the sake of future programmers who don't have as much context as I do at the time of writing, that will be better.

rust/sdk/examples/chat/display.rs
rkuhn: `irssi` has text input only at the bottom; various Unix terminal graphics interfaces place popup boxes in the middle of the screen, e.g. Debian’s package configuration system.

But to really answer this question we’d have to actually do some UI design for this chat application.

rust/sdk/examples/chat/cmdline.rs
Kelerchian: turns out, it doesn't work since one is Some(Identity) and the other is Identity.
I'm keeping this and this https://github.com/Actyx/Actyx/pull/554#discussion_r1360633054

rust/sdk/examples/chat/cmdline.rs
Kelerchian: Keeping this because it is linked to this https://github.com/Actyx/Actyx/pull/554#discussion_r1360634705

rust/sdk/examples/chat/cmdline.rs
rkuhn: You can simply take the Identity out of the Some using `.expect("we know that this is Some")`.

rust/sdk/examples/chat/display.rs
Kelerchian: > various Unix terminal graphics interfaces place popup boxes in the middle of the screen

I see. I am not aware of this culture. No problem then, since this part of the design seems to be more of an interactive process

Kelerchian:
Somehow ratatui's scrollbar doesn't work and I don't see any API that can be used to calculate the height of a rendered paragraph.
For bottom-to-top scrolling, we need a manual calculation of height because `ratatui` expects `scroll` to be in a u16.
Without the required API we can't do the usual method: render-to-fixed-width-buffer=>calculate=>true-render

Curiously it has ScrollDirection that is not used anywhere https://docs.rs/ratatui/latest/ratatui/widgets/enum.ScrollDirection.html

rust/sdk/examples/chat/display.rs
```diff
@@ -127,6 +104,200 @@ pub async fn display(mut cell: ActoCell<Display, impl ActoRuntime>) {
...
+                .style(Style::new().bg(Color::Red)),
+        );
+        f.render_widget(Clear, rect);
+        f.render_widget(text, rect);
+    }
+}
+
+// FIXME sometimes, Scrollbar from ratatui doesn't work for bottom-to-top scrolling
+//  because there is no way to calculate the height of a text
+//  meanwhile Paragraph::wrap has an opaque wrapping mechanism
```
rkuhn, line 246: how about using the `textwrap` crate:
```rust
fn message_to_lines(msg: &Message, width: u16) -> Vec<Line<'static>> {
    let time = DateTime::<Utc>::from(msg.time).to_rfc3339();
    let mut first = vec![
        Span::styled(time, Style::new().add_modifier(Modifier::ITALIC)),
        Span::raw(" "),
        Span::styled(msg.from.to_owned(), Style::new().add_modifier(Modifier::BOLD)),
        Span::raw(": "),
    ];
    let first_width: usize = first.iter().map(|s| s.width()).sum();
    let indent = Span::raw(" ".repeat(first_width));

    let inputs = textwrap::wrap(&msg.text, usize::from(width) - first_width);
    let mut lines = vec![];
    for (idx, input) in inputs.into_iter().enumerate() {
        let mut v = if idx == 0 {
            std::mem::take(&mut first)
        } else {
            vec![indent.clone()]
        };
        v.push(Span::raw(input.into_owned()));
        lines.push(Line::from(v));
    }
    lines
}
```

rust/sdk/examples/chat/display.rs
```diff
@@ -127,6 +104,200 @@ pub async fn display(mut cell: ActoCell<Display, impl ActoRuntime>) {
...
+            // Is there a height shift from previous render?
+            let height_shift_on_update = if let Some(prev_height) = scroll_state.last_recorded_par_height {
+                text_height.checked_sub(prev_height).unwrap_or(0)
+            } else {
+                0
+            };
+
+            scroll_state.last_recorded_par_height = Some(text_height);
+
+            // if vscroll_from_bottom is not 0, it must shift up, because UX
```
rkuhn, line 184: I think the code would become simpler by instead storing the offset from the bottom in the `scroll_state` and only converting that to the `scroll_y` when rendering the Paragraph. Growth of the Paragraph would have to increase the offset unless it is zero.

rust/sdk/examples/chat/display.rs
```diff
@@ -127,6 +104,200 @@ pub async fn display(mut cell: ActoCell<Display, impl ActoRuntime>) {
...
+}
+
+fn render_editing_identity<W: io::Write>(f: &mut Frame<CrosstermBackend<W>>, identity: &Reader<Identity>) {
+    let size = f.size();
+    let pad = (size.height / 2).checked_sub(2).unwrap_or(0);
+    let layout = Layout::default()
+        .direction(ratatui::prelude::Direction::Vertical)
+        .constraints([
+            ratatui::prelude::Constraint::Length(pad),
+            ratatui::prelude::Constraint::Max(3),
```
rkuhn, line 130: This should probably be `Length(3)` since a terminal of less than three lines won’t work anyway.

rust/sdk/examples/chat/display.rs
```diff
@@ -69,56 +80,22 @@ pub async fn display(mut cell: ActoCell<Display, impl ActoRuntime>) {
             Display::Messages(m) => messages = m,
             Display::NotConnected(e) => not_connected = Some(e),
             Display::Connected => not_connected = None,
+            Display::UpdateIdentity(i) => identity = i,
+            Display::Scroll(i) => {
+                let scroll_val = if i >= 0 {
+                    scroll_state
+                        .vscroll_from_bottom
+                        .checked_add(i as u16)
+                        .unwrap_or(u16::MAX)
```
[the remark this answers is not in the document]
Kelerchian: Ah, I was misled to think that saturating_x is a nightly feature

rust/sdk/examples/chat/display.rs
Kelerchian: Ah that is right.

rust/sdk/examples/chat/cmdline.rs
```diff
@@ -12,40 +12,138 @@ pub enum Cmdline {
...
+                                    -1 * {
+                                        if key.modifiers.contains(KeyModifiers::CONTROL) {
+                                            10
+                                        } else {
+                                            1
+                                        }
+                                    },
+                                ));
+                            }
+                            KeyCode::Up => {
```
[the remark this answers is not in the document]
Kelerchian: Alright. This, I think you will know more than I do since I rarely use interactive terminals

rust/sdk/examples/chat/display.rs
Kelerchian: In the rare case of a very long username, a very narrow terminal, or a very deep zoom, won't that break the bold modifier of the username if it is broken in the middle of the username?

rust/sdk/examples/chat/display.rs
Kelerchian: That is now what the code does. Although it might have a little bit confusing

Kelerchian: Also, my comment is misleading

rust/sdk/examples/chat/cmdline.rs
```diff
@@ -12,40 +12,138 @@ pub enum Cmdline {
...
+                                        edit.input(key);
+                                    }
+                                }
+                            }
+                        }
+                        display.send(Display::UpdateIdentity(identity.reader()));
+                    }
+                    // When chatting
+                    false => {
+                        match key.code {
```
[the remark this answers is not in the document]
Kelerchian: I am not sure about spawning another actor because then we will indicate that somehow identity controller can potentially move non-synchronously with the rest of the cmdline.
I am also not sure about `input` recognizing `page up` and `page down` because at least the model in my head says precisely what you have written: "Input is alive and running unless there is a CTRL+C"

However, I think it is good to separate the "identity" mode as a "sub-state-machine" of cmdline's state machine. But this is done for code organization and not software architecture.

rust/sdk/examples/chat/display.rs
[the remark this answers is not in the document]
Kelerchian: fixed in https://github.com/Actyx/Actyx/pull/554/commits/a52d3233b2e80053dbe04d872393d0b23e2aa816

rust/sdk/examples/chat/cmdline.rs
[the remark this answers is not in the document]
Kelerchian: fixed in https://github.com/Actyx/Actyx/pull/554/commits/a52d3233b2e80053dbe04d872393d0b23e2aa816

rust/sdk/examples/chat/display.rs
Kelerchian: made more readable in https://github.com/Actyx/Actyx/pull/554/commits/a52d3233b2e80053dbe04d872393d0b23e2aa816

rust/sdk/examples/chat/display.rs
Kelerchian: should be clearer in https://github.com/Actyx/Actyx/pull/554/commits/a52d3233b2e80053dbe04d872393d0b23e2aa816

rust/sdk/examples/chat/display.rs
Kelerchian: fixed in https://github.com/Actyx/Actyx/pull/554/commits/a52d3233b2e80053dbe04d872393d0b23e2aa816

rust/sdk/examples/chat/display.rs
```diff
@@ -52,18 +52,46 @@ pub enum Display {
...
 }
 
+impl MessageScrollState {
+    pub fn update_text_height_and_clamp_vscroll(&mut self, text_height: u16, clamp_max_vscroll: u16) {
+        // if scroll_state.vscroll_from_bottom is not 0, it must shift up, because UX
+        if self.vscroll_from_bottom > 0 {
+            // Is there a height shift from previous render?
+            let height_shift_from_previous_render = self
+                .last_text_height
+                .map_or(0_u16, |prev_height| text_height.saturating_sub(prev_height));
```
rkuhn, line 83: This means that if the Paragraph got smaller, the scroll offset from bottom stays the same — which is correct if we assume that getting smaller means messages being dropped at the top. If it gets smaller because the terminal width is increased, then we get an effective “scrolling upwards” behaviour. This may be fine, I just wanted to say it.

rust/sdk/examples/chat/cmdline.rs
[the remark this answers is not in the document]
rkuhn: The identity edit field is not used concurrently with the message composer (which I somehow misnamed `cmdline`), either one has focus or the other. In the future we may have other things focused, so I do think we need to have a focus handler. This handler cannot be inside `cmdline` because `cmdline` can go out of focus itself, hence my plan to place this code inside `input` — after all, it is about routing input events to the right places.

I also think PageUp/PageDown have nothing to do with `cmdline`, they are only of interest to `input` and `display`.

rust/sdk/examples/chat/display.rs
rkuhn: The main point of using `textwrap` is that it handles the fact that not every byte corresponds to a glyph on the screen, and it breaks nicely at whitespace. The other point is to indent all lines after the first. With the current setup (i.e. scroll_x is always zero) a very narrow terminal only leads to not seeing the message at all, which I think is fine. The current format of the timestamp is extremely technical and should probably be shortened to a more usual appearance, but for a demo it is good enough.

rust/sdk/examples/chat/display.rs
rkuhn: yes, it is

rust/sdk/examples/chat/display.rs
Kelerchian: Ah, yeah. I did not catch that scenario

rkuhn (approved):

PR #557, opened 2023-10-16
Improve `ActyxClient::new`

### LINKED ISSUE #552: `ActyxClient::new` is too verbose
Example:
```rust
let client = ActyxClient::new(
        "http://localhost:4454".parse().unwrap(),
        AppManifest::new(
            app_id!("com.example.chat"),
            "simple chat".to_owned(),
            "0.1.0".to_owned(),
            None,
        ),
    )
    .await?;
```

A possible solution would be to offer something like:
```rust
// "http://localhost:4454" is assumed as a default connection location
let client = ActyxClient::local(
    AppManifest::new(...)
).await?;
```

However, this approach doesn't scale with future options, so another possibility could be 
```rust
let client = ActyxClient::new(ActyxClientOptions {
    manifest: AppManifest::new(),
    Default::default..
}).await?;
```

Furthermore, we can reduce it further by providing a `Default` implementation for `AppManifest` that returns a trial manifest (if the user wants to rename the manifest they can do so by using the "normal path").
For reference, said manifest would be something like:
```rust
AppManifest {
  app_id: app_id!("com.example.trial"),
  "trial_app",
  "0.0.1",
  None
}
```

Finally, we can also reduce verbosity on `AppManifest` by taking `&str` instead of `String` and cloning inside the method instead.

jmg-duarte:
closes #552

### DIFF
--- a/rust/actyx/node/benches/local_event_roundtrip.rs
+++ b/rust/actyx/node/benches/local_event_roundtrip.rs
@@ -1,26 +1,13 @@
 use actyx_sdk::{
-    app_id,
     service::{EventService, Order, PublishEvent, PublishRequest, QueryRequest},
-    tags, ActyxClient, AppManifest, Payload,
+    tags, ActyxClient, Payload,
 };
 use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use futures::StreamExt;
 use node::{BindTo, Runtime};
 use std::time::Duration;
 use tempfile::tempdir;
-use url::Url;
 use util::SocketAddrHelper;
-
-async fn mk_http_client() -> anyhow::Result<ActyxClient> {
-    let app_manifest = AppManifest::new(
-        app_id!("com.example.trial-mode"),
-        "display name".into(),
-        "0.1.0".into(),
-        None,
-    );
-    let url = Url::parse("http://localhost:4454").unwrap();
-    ActyxClient::new(url, app_manifest).await
-}
 
 // Note: This doesn't concern itself with any internals (like flushing the send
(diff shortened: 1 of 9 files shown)


rust/sdk/src/app_manifest.rs
```diff
@@ -26,3 +26,19 @@ impl AppManifest {
         }
     }
 }
+
+impl Default for AppManifest {
+    /// Returns the default manifest.
+    ///
+    /// The default manifest has the application ID `com.example.trial`,
+    /// the display name "Trial App" and the version "0.0.1".
+    /// Since it is a trial applicatio manifest, it does not have a signature.
```
rkuhn, line 35:
```suggestion
    /// Since it is a trial application manifest, it does not have a signature.
```

rust/actyx/node/benches/local_event_roundtrip.rs
```diff
@@ -49,7 +36,7 @@ fn round_trip(c: &mut Criterion) {
     }
     c.bench_function("id", |b| {
         b.to_async(&rt).iter_batched(
-            || (data.clone(), mk_http_client()),
+            || (data.clone(), ActyxClient::new(AxClientOpts::default())),
```
rkuhn, line 39: How about `Ax::client(AxOpts::default())`?

jmg-duarte: I wasn't changing the names for the client yet. I can do it though

Kelerchian (approved):

rkuhn (approved):

PR #560, opened 2023-10-18
Implement a builder for the `ActyxClient`

### LINKED ISSUE #553: `ActyxClient` subscribe methods are too verbose
There's a bunch of reasons the API is verbose:
- Requires `String`s instead of `&str` (taking ownership inside the function)
- Methods require structures with the same name as the method, leading to the repetition of names through out

Example:
```rust
client
    .subscribe_monotonic(SubscribeMonotonicRequest {
        session: SessionId::from("me"),
        query: "FROM appId(com.example.chat) & 'message'".to_owned(),
        from: StartFrom::LowerBound(OffsetMap::empty()),
    })
    .await
    .expect("failed to subscribe to messages")
```

From here, the only _truly_ required field is `query`.

One option is changing the `subscribe_monotonic` to receive just `query` instead:
```rust
client
    .subscribe_monotonic("FROM appId(com.example.chat) & 'message'")
    .await
    .expect("failed to subscribe to messages")
```
This would initialize the request inside the method instead of requiring the user to do so.

Another option would be to use `Default` and allow the user to write:
```rust
client
    .subscribe_monotonic(SubscribeMonotonicRequest {
        query: "FROM appId(com.example.chat) & 'message'".to_owned(),
        Default::default()..
    })
    .await
    .expect("failed to subscribe to messages")
```

Yet another (more generic) option would be to create a public `request` method, that receives these `Request` objects:
```rust
enum Request { SubscribeMonotonic(SubscribeMonotonicRequest), ... }
fn request(request: Request) {
    match request {
        SubscribeMonotonic(request) => { ... }
    }
}
```
We could even make the `SubscribeMonotonic` inside the enumeration a structure instead of an anonymous tuple (apologies if the name is wrong). Serialization could be made `untagged` so the underlying representation stays the same too.

### LINKED ISSUE #555: `ActyxClient::publish` is too verbose
```rust
client
    .publish(PublishRequest {
        data: vec![PublishEvent {
            tags: tags!("message"),
            payload: Payload::compact(&event).expect("failed to serialize event"),
        }],
    })
    .await
    .expect("failed to publish event");
```

This function suffers from the same issue as https://github.com/Actyx/Actyx/issues/553.

- We can provide a function for a single event, or multiple events.
- Publishing several events with the same tags is also not ergonomic — the user is required to write down the tags multiple times. 

Some like the following would be more ergonomic:
```rust
client.publish_events(&tags, vec![event1, event2])
```

- The `payload` detail could also be hidden from the user, just forcing the type that goes into the function to be `T: Serializable` to match the `compact` constraint. While the user may construct a payload from a slice, I'm not expecting that to be the "normal case", currently, we use `Payload::from_slice` in just two places: https://sourcegraph.com/search?q=context%3Aglobal+repo%3Agithub.com%2FActyx%2FActyx+Payload%3A%3Afrom_slice&patternType=standard&sm=1&groupBy=repo

For example:
```rust
client
    .publish_event(tags!("hello", "there"), &event)
```

And we could still keep `publish` as a more low-level abstraction (still keeping the `Payload::compact` away from the user):
```rust
client.publish(vec![PublishEvent::new(tags!("hey", "there"), &event)])
```

jmg-duarte:
closes #553
closes #555 

Still WIPish 
- [x] Docs
- [x] `subscribe` example
- [x] `subscribe_monotonic` example
- [x] Tests (harness like)

Is also waiting on #557


rust/sdk/src/client.rs
```diff
@@ -282,86 +263,33 @@ impl EventService for ActyxClient {
...
-            .await
-            .context(|| format!("getting body for GET {}", self.events_url("publish")))?;
-        Ok(serde_json::from_slice(bytes.as_ref()).context(|| {
-            format!(
-                "deserializing publish response from {:?} received from GET {}",
-                bytes,
-                self.events_url("publish")
-            )
-        })?)
+    /// Returns a builder for publishing events.
```
rkuhn, line 266: Please add an example showing that the Builder also is a Future, i.e. can be `.await`ed. I’d use a single event to demo it because that will be the most frequent usage.

rust/sdk/src/client.rs
```diff
@@ -282,86 +263,33 @@ impl EventService for ActyxClient {
...
-        let body = serde_json::to_value(&request).context(|| format!("serializing {:?}", &request))?;
-        let response = self
-            .do_request(|c| c.post(self.events_url("query")).json(&body))
-            .await?;
-        let res = to_lines(response.bytes_stream())
-            .map(|bs| serde_json::from_slice(bs.as_ref()))
-            // FIXME this swallows deserialization errors, silently dropping event envelopes
-            .filter_map(|res| future::ready(res.ok()));
-        Ok(res.boxed())
+    /// Returns a builder to query events.
```
rkuhn, line 271: similar here, showing two queries (one directly awaiting, the other with setting some option)

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    /// Add a lower bound to the query.
+    ///
+    /// The lower bound limits the start of the query events.
+    /// As an example, consider the following (example) events:
+    /// ```json
+    /// { "lamport": 1, "event": { "temperature": 10 } }
+    /// { "lamport": 3, "event": { "temperature": 12 } }
+    /// { "lamport": 14, "event": { "temperature": 9 } }
+    /// ```
+    /// If you set the lower bound to `10`, only the last event will be returned.
```
rkuhn, line 432: This explanation is incorrect: the lower bound uses Offsets, not timestamps, so the interpretation isn’t that easy. It is also impossible to just make up an OffsetMap because you’d need some StreamIds that you cannot guess. An example for this method would need to involve a call to `client.offsets()` to get an OffsetMap, I think.

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    /// { "lamport": 3, "event": { "temperature": 12 } }
+    /// { "lamport": 14, "event": { "temperature": 9 } }
+    /// ```
+    /// If you set the upper bound to `10`, the first two events will be returned.
+    pub fn with_upper_bound(mut self, upper_bound: OffsetMap) -> Self {
+        if let Self::Initial { ref mut request, .. } = self {
+            request.upper_bound = Some(upper_bound);
+        }
+        self
+    }
```
rkuhn, line 455: Since the default here means “present” and can thus not easily be represented using `.with_upper_bound()`, I think there should be a `.without_upper_bound()` companion.

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    /// If your query sets [`Order::Desc`], the result will instead look like:
+    /// ```json
+    /// { "lamport": 14, "event": { "temperature": 9 } }
+    /// { "lamport": 3, "event": { "temperature": 12 } }
+    /// { "lamport": 1, "event": { "temperature": 10 } }
+    /// ```
+    pub fn with_order(mut self, order: Order) -> Self {
+        if let Self::Initial { ref mut request, .. } = self {
+            request.order = order;
+        }
```
rkuhn, line 474: I’m on the fence, but slightly leaning towards panicking in the `else` branch.

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+        let this = self.get_mut();
+        loop {
+            *this = match replace(this, Query::Empty) {
+                Query::Initial { client, request } => {
+                    let query_response = async move {
+                        let query_url = client.events_url("query");
+                        let response = client.do_request(|c| c.post(query_url).json(&request)).await?;
+                        let response_stream = to_lines(response.bytes_stream())
+                            .map(|bytes| serde_json::from_slice::<QueryResponse>(&bytes))
+                            // FIXME this swallows deserialization errors, silently dropping event envelopes
```
rkuhn, line 492: We should probably use `tracing` in the SDK in general, and here we should log the error — dropping the bytes from the stream is still the only reasonable action to take here

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+            client,
+            request: SubscribeMonotonicRequest {
+                query: query.into(),
+                session: SessionId::from("me"),
+                from: StartFrom::LowerBound(OffsetMap::empty()),
+            },
+        }
+    }
+
+    // TODO: Figure out how to explain this
```
rkuhn, line 605: we should probably leave this one off for now, since the session id isn’t used for anything

rust/sdk/src/client.rs
jmg-duarte: I was going to write a full one, but since we didn't merge #557 yet, I didn't to avoid duplicate work 😅

rust/sdk/src/client.rs
jmg-duarte: How could it happen? 🤔 Could someone be messing with the internals?
Like on the `Empty` case from the future, I prefer crashing than silently failing

rust/sdk/src/client.rs
jmg-duarte: I'll take the `pub` out then, to avoid breaking changes.

rust/sdk/src/client.rs
rkuhn: Well, someone would have to

- create the Future
- manually poll it once
- then call such a method

This is so clearly an error that a panic seems justified. These Futures should be consumed using `.await`, in which case no such problem can occur.

rust/sdk/examples/publish_stream.rs
```diff
@@ -0,0 +1,40 @@
...
+                PublishEvent {
+                    tags: tags!("com.actyx.examples.temperature", "sensor:temp-sensor1"),
+                    payload: Payload::compact(&serde_json::json!({ "counter": i })).unwrap(),
+                },
+                PublishEvent {
+                    tags: tags!("com.actyx.examples.temperature", "sensor:temp-sensor2"),
+                    payload: Payload::compact(&serde_json::json!({ "counter": i })).unwrap(),
+                },
+            ]
+            .into_iter(),
```
Kelerchian, line 31: Is this `into_iter` still needed? (since events receives `E: IntoIterator<_>`)

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    }
+}
+
+pub enum Query<'a> {
+    Initial {
+        client: &'a ActyxClient,
+        request: QueryRequest,
+    },
+    Pending(BoxFuture<'a, anyhow::Result<BoxStream<'a, QueryResponse>>>),
+    Empty,
```
Kelerchian, line 407: A nitpick:
`Spent`, `Depleted`, or `Exhausted` can give more meaning instead of `Empty`.
My initial reaction to Empty is that it felt like the initial state rather than the end.

rust/sdk/src/client.rs
Kelerchian: Should this be already implied by the non-calling of `with_upper_bound`?

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    /// { "lamport": 3, "event": { "temperature": 12 } }
+    /// { "lamport": 14, "event": { "temperature": 9 } }
+    /// ```
+    /// If your query sets [`Order::Desc`], the result will instead look like:
+    /// ```json
+    /// { "lamport": 14, "event": { "temperature": 9 } }
+    /// { "lamport": 3, "event": { "temperature": 12 } }
+    /// { "lamport": 1, "event": { "temperature": 10 } }
+    /// ```
+    pub fn with_order(mut self, order: Order) -> Self {
```
Kelerchian, line 471: What happens if I pass this query `FROM 'sometag' ORDER DESC` and then set `with_order(Order::ASC)`?

rust/sdk/src/client.rs
```diff
@@ -388,6 +316,356 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+        let this = self.get_mut();
+        loop {
+            *this = match replace(this, Self::Empty) {
+                Self::Initial { client, request } => {
+                    let query_response = async move {
+                        let query_url = client.events_url("subscribe");
+                        let response = client.do_request(|c| c.post(query_url).json(&request)).await?;
+                        let response_stream = to_lines(response.bytes_stream())
+                            .map(|bytes| serde_json::from_slice::<SubscribeResponse>(&bytes))
+                            // FIXME this swallows deserialization errors, silently dropping event envelopes
```
Kelerchian, line 564: I think this should be addressed now rather than later simply because it may probably cause a change in the `Output` type.

rust/sdk/src/client.rs
rkuhn: https://github.com/Actyx/Actyx/blob/7685fe6d1fba2f577cc804b3b60831d71d843418/rust/actyx/api/src/events/service.rs#L139

The precedence is:
1. from AQL
2. inferred through AGGREGATE
3. from the API request

rust/sdk/src/client.rs
rkuhn: Since this is also used as an intermediate value while polling, `Void` would be more appropriate.

rust/sdk/src/client.rs
jmg-duarte: Great question, I've taken this into account for the next revision

rust/sdk/examples/publish_stream.rs
jmg-duarte: It isn't. I just didn't push everything from yesterday.

The new batch comes with a bunch of new examples and revisions.

Kelerchian: I see. No problems, just making sure that you don't miss these invisible things

rust/sdk/src/client.rs
Kelerchian: Ah, I see. I don't know that that pattern is a thing

rust/sdk/src/client.rs
jmg-duarte: I think we shouldn't address this in _this_ PR, it's already massive. I can open an issue to investigate this afterwards

rust/sdk/src/client.rs
Kelerchian: Should we receive offsets from Actyx API exclusively?

rust/actyx/swarm/harness/src/bin/query.rs
```diff
@@ -44,24 +30,28 @@ fn main() -> anyhow::Result<()> {
...
-        let api = Api::new(&mut sim, app_manifest)?;
+        let api = Api::new(&mut sim, AppManifest::default())?;
 
-        for machine in sim.machines() {
+        for (idx, machine) in sim.machines().iter().enumerate() {
+            tracing::error!("{}", idx);
             api.run(machine.id(), |api| async move {
-                api.publish(PublishRequest { data: make_events(N) }).await?;
+                api.0
+                    .spawn_mut(|ax| {
```
Kelerchian, line 39: Curious, what does spawn_mut + block_on achieve compared to plainly calling `ax.publish()`?

rust/actyx/swarm/harness/src/bin/versions.rs
```diff
@@ -116,133 +115,150 @@ fn main() -> anyhow::Result<()> {
                     }
                     Err(e) => {
                         tracing::info!("{} not yet ready: {}", i, e);
-                        sleep(Duration::from_secs(1));
+                        async_std::task::sleep(Duration::from_secs(1)).await;
```
Kelerchian, line 118: Why use `async_std` instead of `tokio` here?
And if it is because something underneath machine uses `async_std`, is there any guide to where to use `async_std`'s sleep and where to use `tokio`'s sleep?

rust/actyx/swarm/harness/src/bin/query.rs
jmg-duarte: `ax` is behind `PinnedResource`:
```rust
/// Protects a resource so that it's always tied to a single, dedicated thread. This wrapper allows other
/// places to safely access the underlying resource even if that is not Sync or even Send. This
/// is achieved by having a separate, dedicated thread hosting the resource which then communicates
/// via a message queue with the rest of the system.
///
/// This wrapper is also useful to isolate a blocking resource from a thread-pool that hosts other
/// async tasks.
pub struct PinnedResource<T> {
    stopped: Arc<()>, // Just for keeping track of the count of external handles
    sender: Sender<Box<dyn FnOnce(&mut T) + Send>>,
}
```
From my interpretation, this is a way to keep accesses to the API orderly while changing machines and so on during the tests.
I've kept it as similar as possible to reduce entropy and to be able to deliver the PR, otherwise I'm guessing I'd still be investigating more things.

rust/actyx/swarm/harness/src/bin/versions.rs
jmg-duarte: I've used `async_std` given the rest of the tests mainly use `async_std`, such as the `block_on`. Ideally, we'd move everything to one of them, for now, this is how I tried to cope 😅

rust/actyx/swarm/harness/src/bin/query.rs
rkuhn: IIRC this also ensures that the API calls are performed from a thread that is in the correct network namespace to talk to the Actyx instance.

rust/actyx/swarm/harness/src/bin/quickcheck_ephemeral.rs
```diff
@@ -142,7 +142,7 @@ fn main() {
             let mut present = OffsetMap::empty();
             let machine = sim.machines().first().unwrap();
             api.run(machine.id(), move |client| async move {
-                client.publish(PublishRequest { data: events }).await?;
+                client.0.spawn_mut(|ax| block_on(ax.publish().events(events))).await??;
```
rkuhn, line 145: instead of making `.0` public, how about a `client.exec(...)` method?

rust/sdk/src/client.rs
```diff
@@ -457,6 +458,684 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+            request: QueryRequest {
+                query: query.into(),
+                lower_bound: Some(OffsetMap::empty()),
+                upper_bound: None,
+                order: Order::Asc,
+            },
+        }
+    }
+
+    /// Add a (inclusive) lower bound to the query.
```
rkuhn, line 607: the lower bound is exclusive (and the upper bound inclusive) because that is much simpler to track than the other way around: the lower bound defines what you have already seen and don’t want to see again

rkuhn (commented):
not yet completely reviewed, will return

rust/actyx/swarm/harness/src/bin/quickcheck_ephemeral.rs
jmg-duarte: Sure but let's do it in another issue/PR, this PR is getting too complex

rust/sdk/src/client.rs
jmg-duarte: Indeed, I had an extra print in my test example and got confused. Nice catch!

rust/sdk/src/client.rs
```diff
@@ -457,6 +458,684 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    fn new(client: &'a Ax) -> Self {
+        Self::Initial {
+            client,
+            request: PublishRequest { data: vec![] },
+        }
+    }
+
+    /// Add an event.
+    ///
+    /// Subsequent calls to this function will not remove previous events.
```
rkuhn, line 481: This is close to a double negation. How about “This adds the given event to the list of events that will be emitted once this publishing request is submitted by using `.await` on it.”

rust/sdk/src/client.rs
```diff
@@ -457,6 +458,684 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    /// ```
+    pub fn event<E: Serialize>(mut self, tags: TagSet, event: &E) -> Result<Self, serde_cbor::Error> {
+        if let Self::Initial { ref mut request, .. } = self {
+            request.data.push(PublishEvent {
+                tags,
+                payload: Payload::compact(event)?,
+            });
+            return Ok(self);
+        }
+        panic!("Calling Publish::event after polling.")
```
rkuhn, line 512: All these panics need to be visible in the respective method’s doc comment to help people not making the mistake.

rust/sdk/src/client.rs
```diff
@@ -457,6 +458,684 @@ pub(crate) fn to_lines(stream: impl Stream<Item = Result<Bytes, reqwest::Error>>
...
+    /// ```
+    pub fn with_upper_bound(mut self, upper_bound: OffsetMap) -> Self {
+        if let Self::Initial { ref mut request, .. } = self {
+            request.upper_bound = Some(upper_bound);
+            return self;
+        }
+        panic!("Calling Query::with_upper_bound after polling.")
+    }
+
+    /// Dual to [`Query::with_upper_bound`], removes the upper bound.
```
rkuhn, line 793: Please document what this means: the upper bound will be filled in by Actyx when processing the query, to be the currently known “present” offsets.

rkuhn (approved):
LGTM, with some minor fixes.

rust/sdk/src/client.rs
jmg-duarte: Done so according to https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure

PR #565, opened 2023-10-20
Refactor harness/versions for readability / maintainability

jmg-duarte:


rust/actyx/swarm/harness/src/bin/versions.rs
```diff
@@ -149,100 +159,106 @@ fn main() -> anyhow::Result<()> {
...
+            &streams,
+            "versions",
+            Duration::from_secs(20),
+            Duration::from_secs(1),
+        )
+        .await?;
+
+        // it may be that the above was fulfilled by indirect event delivery, but now every
+        // node must see every other; just wait a second to get them connected
+        async_std::task::sleep(Duration::from_secs(1)).await;
```
Kelerchian, line 230: curious, is this a similar attempt to fully_meshed?

rust/actyx/swarm/harness/src/bin/versions.rs
```diff
@@ -94,52 +86,70 @@ fn main() -> anyhow::Result<()> {
...
+    async fn ensure_all_machines_are_up(
+        sim: &mut Netsim<Command, Event>,
+        machine_ids: &Vec<MachineId>,
+    ) -> anyhow::Result<()> {
+        for i in machine_ids {
             let ip = sim.machine(*i).addr();
             let ns = sim.machine(*i).namespace();
             loop {
-                let req = std::thread::spawn(move || {
+                let req = {
```
rkuhn, line 100: This changes semantics: `ns.enter()` places the current thread into the `ns` namespace, hence the new thread.

Kelerchian (commented):
looking good

rust/actyx/swarm/harness/src/bin/versions.rs
jmg-duarte: Good question, I would need to read it to know 😅

rust/actyx/swarm/harness/src/bin/versions.rs
```diff
@@ -149,100 +159,106 @@ fn main() -> anyhow::Result<()> {
...
+        // These values date back to before the refactor
+        publish_and_query(
+            &api,
+            &machines,
+            &streams,
+            "version2",
+            Duration::from_secs(1),
+            Duration::from_millis(100),
+        )
+        .await?;
```
rkuhn, line 243: Yes, this structure is clearer.

rust/actyx/swarm/harness/src/bin/versions.rs
jmg-duarte: I suspected something like that was happening. So I guess that this works instead:
```rust
let req = timeout(Duration::from_secs(10), async { 
    ns.enter().unwrap();
    reqwest::get(format!("http://{}:4454/api/v2/node/id", ip))
}).await;
```

rkuhn: no: now you put some arbitrary tokio thread into that namespace …

jmg-duarte: Ah right, the thread doesn't leave the namespace when finished. I'll get the thread back

jmg-duarte:
Closing this in favor of #560

PR #566, opened 2023-10-20
add warning to multiple sdk instances with the same parameters

### LINKED ISSUE #564: [Feature]: Print warning when more than 1 SDK is created in a NodeJS/Browser runtime
### Product

Actyx SDK

### When

More than one SDK instance is created (e.g. `Actyx.of`) for the same manifest

### I want to

be warned of that

### So that

I can make sure only 1 instance is active at a time

### Additional notes

_No response_

Kelerchian:


js/sdk/src/actyx.ts
```diff
@@ -163,3 +188,60 @@ export const Actyx = {
...
+/**
+ * "Reference counter" for living Actyx
+ */
+const actyxInstanceRegister = (() => {
+  type AppId = AppManifest['appId']
+  type Version = AppManifest['version']
+  type Signature = AppManifest['signature']
+  type ActyxHost = ActyxOpts['actyxHost']
+  type ActyxPort = ActyxOpts['actyxPort']
+  type Param = [AppId, Version, Signature, ActyxHost, ActyxPort]
```
jmg-duarte, line 201: I dislike the `Param` name, `Params`/`Parameters` makes much more sense given it's a list.

I'd also add a brief docstring

jmg-duarte (commented):
If I understand correctly, this provides two "entry points" to the Actyx client. If that is the case, I feel that the behavior should be better documented for the user.

Kelerchian:
> If I understand correctly, this provides two "entry points" to the Actyx client. If that is the case, I feel that the behavior should be better documented for the user.

No. This is purely an internal logging mechanism opaque to the user

js/sdk/src/actyx.ts
```diff
@@ -123,18 +123,43 @@ export const Actyx = {
...
+        JSON.stringify(manifest),
+      )
+      return createV2(manifest, opts, nodeId)
+    })()
+
+    // Patch dispose
+    const oldDispose = actyx.dispose
+    actyx.dispose = () => {
+      const result = oldDispose()
+      actyxInstanceRegister.subOne(manifest, opts)
```
rkuhn, line 157: Does `oldDispose()` invoke some user callbacks? If so, such callback could try to create a new SDK instance, which in that case would log the “multiple connections” error. If correctness permits it, I’d do the `subOne` before the `oldDispose`.

js/sdk/src/actyx.ts
rkuhn: these types aren’t visible to anyone, so a docstring won’t be displayed anywhere — did you mean a comment?

js/sdk/src/actyx.ts
```diff
@@ -123,18 +123,43 @@ export const Actyx = {
...
-    if (!nodeId) {
-      // Try connecting to v1 if we failed to retrieve a v2 node id
-      // (Note that if the port is completely unreachable, v2getNodeId will throw an exception and we don’t get here.)
-      log.actyx.debug('NodeId was null, trying to reach V1 backend...')
-      return createV1(opts)
+    const previousInstanceCount = actyxInstanceRegister.count(manifest, opts)
+    if (previousInstanceCount > 0) {
+      console.warn(
+        new Error(
+          `Multiple creations of actyx instance with these parameters are detected! manifest: ${JSON.stringify(
```
rkuhn, line 130:
```suggestion
          `Multiple connections to Actyx with these parameters are detected! manifest: ${JSON.stringify(
```

js/sdk/src/actyx.ts
Kelerchian: It only calls `MultiplexedWebsocket.dispose` under the hood.
There should be no difference between both orders since this is a sync function.

rkuhn: My worry was that there might be an event emitted to user code from that sync function — and if we add such a thing in the future then this problem may well surface.

Kelerchian: Ah, that's a valid concern

Kelerchian: fixed https://github.com/Actyx/Actyx/pull/566/commits/fc39630f998fe973970932f29afb1bf9e89934e4

PR #568, opened 2023-10-24
move app_manifest to sdk

Kelerchian:
A draft. Not done yet: move tests to proper place and tweak them to have proper specs


Kelerchian:
Moving AppManifest to sdk/app_manifest.
This changes the AppManifest creation slightly

jmg-duarte (approved):
LGTM

Not a huge fan of the `internals` feature but I don't have any better suggestions

rust/actyx/certs/src/app_manifest.rs
```diff
@@ -103,65 +67,34 @@ impl AppManifest {
...
-            display_name,
-            version,
-            signature: Some(manifest_signature),
-        }))
-    }
-
-    pub fn signed(app_id: AppId, display_name: String, version: String, signature: AppManifestSignature) -> Self {
-        Self(AppManifestIo {
+        let manifest_signature_string: String = manifest_signature.try_into()?;
+        let manifest: AppManifest = AppManifestIo {
```
rkuhn, line 71: Why use `AppManifestIo` here instead of just creating an `AppManifest`? That would also make the `internal` feature unnecessary.

rkuhn (commented):
Looking good, apart from the leakage of `AppManifestIo`.

rust/actyx/certs/src/app_manifest.rs
Kelerchian: If we remove AppManifestIo, does the `TryInto` stuff go to the deserializer?

rkuhn: Not sure I understand the question: what I mean is that `AppManifestIo` is (and shall remain) a purely internal detail of how `AppManifest` is (de)serialized. So just remove the export and the `internals` feature and directly construct an `AppManifest` here.

Kelerchian: Ah, I see.
Currently, `signed` manifest is purely created in the Actyx (by constructing `AppManifestIo` manually), no API is available to the user for creating a `signed` manifest. That's why the `internals` bridge.

PR #569, opened 2023-10-25
fix stopped node reconnections

Kelerchian:
Node manager stops trying to connect at some point and there seems to be no "reconnect" button.
This PR change the behavior so that node manager doesn't stop attempting to connect


rkuhn (approved):
The old code structure employed a function for each round of tries — maybe there was some error handling missing, but I still think that that is better than having a single asynchronous `while` loop: any thrown exception will terminate the connection machinery.

PR #573, opened 2023-10-26
Add subtitle explanining Swarm Settings

### LINKED ISSUE #571: Add "subtitle" to Swarm Settings explaning that changes are applied to connected nodes only

jmg-duarte:
closes #571 

Maybe we can also apply this subtitle pattern to other pages

### DIFF
--- a/js/node-manager/src/client/components/Layout.tsx
+++ b/js/node-manager/src/client/components/Layout.tsx
@@ -199,6 +199,7 @@ export const Menu: React.FC<MenuProps & ClassName> = ({ className, items }) => {
 }
 interface LayoutProps {
   title?: string
+  subtitle?: string
   input?: InputProps
   actions?: ActionProps[]
   menuItems?: MenuItem[]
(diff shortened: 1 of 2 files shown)


js/node-manager/src/client/components/Layout.tsx
```diff
@@ -316,9 +318,16 @@ export const Layout: React.FC<LayoutProps & ClassName> = ({
       </div>
       <div className="flex flex-col flex-grow overflow-hidden">
         <div className="flex items-center flex-shrink-0 h-16 px-7 border-b border-gray-300 z-10">
-          <h1 className="text-lg font-medium" id="Layout_Title">
-            {title}
-          </h1>
+          <div className="flex flex-col">
```
Kelerchian, line 321: [Non-major] Does changing between page make the text go up and down?
If so, then add `justify-start py-2` here

Kelerchian, line 321: [Non-major] Does changing between page make the text go up and down?
If so, then add `justify-start py-2` here

Kelerchian (approved):
Looks good except 1 item

PR #577, opened 2023-11-01
reroute swarm cli log to file

Kelerchian:


rust/actyx/swarm/cli/src/main.rs
```diff
@@ -25,25 +25,50 @@ use tokio::{
...
 };
+use tracing_subscriber::{fmt::format::FmtSpan, EnvFilter};
 use trees::{query::TagExprQuery, AxKey};
 use util::variable::Writer;
 
+fn make_log_filename() -> String {
+    std::env::vars()
+        .find(|x| x.0 == "NETSIM_TEST_LOGFILE")
+        .map(|x| x.1)
+        .unwrap_or("unknown".to_string())
```
rkuhn, line 36: this could be simplified to `std::env:var("NETSIM_TEST_LOGFILE").unwrap_or_else(|_| "unknown".to_owned())`

rkuhn (commented):
This should be closed because it was rebased, right? But still I found one small thing to improve :-)

rust/actyx/swarm/cli/src/main.rs
Kelerchian: I'll include the fix on my next PR

PR #578, opened 2023-11-01
simplify connected-disconnected detection

Kelerchian:


rust/actyx/swarm/cli/src/main.rs
```diff
@@ -126,19 +124,31 @@ async fn run() -> Result<()> {
...
-                if is_connected && !connected_peer_ids.contains(&peer_id) {
-                    println!("{}", Event::Connected(peer_id));
-                    connected_peer_ids.insert(peer_id);
-                }
+        let mut last_connected = HashSet::<PeerId>::new();
+        loop {
+            let current_connected = ipfs
+                .connections()
+                .iter()
+                .filter(|connection| ipfs.is_connected(&connection.0))
```
rkuhn, line 135: does this filter ever reject an entry? I’d think that the presence of a connection indicates that we are connected

rust/actyx/swarm/cli/src/main.rs
```diff
@@ -126,19 +124,31 @@ async fn run() -> Result<()> {
...
+                .map(|connection| connection.0)
+                .collect::<HashSet<PeerId>>();
+
+            let new_connected = &current_connected - &last_connected;
+            let new_disconnected = &last_connected - &current_connected;
+
+            new_connected.iter().for_each(|peer_id| {
+                println!("{}", Event::Connected(peer_id.clone()));
```
rkuhn, line 143: PeerId is copy, so no need for cloning; but I’d use `into_iter` instead of `iter` on the line above

rkuhn (approved):

rust/actyx/swarm/cli/src/main.rs
Kelerchian: Ah you're right.

PR #583, opened 2023-11-08
Remove the StartFrom abstraction and add doc to SubscribeMonotonic

### LINKED ISSUE #582: Handle missing doc for `start_from` in the Rust's SDK `SubscribeMonotonic`
What the title said

jmg-duarte:
closes #582 

There's still info missing and I think the bounds examples may be poorly documented because querying for a specific event in the middle of something doesn't yield the last offset of that event, it still sends the present; as seen in: https://github.com/Actyx/Actyx/blob/3c85d303451d0e9e20404c06742e1fc323751aef/rust/actyx/api/src/events/service.rs#L122-L125

### DIFF
--- a/rust/sdk/src/client.rs
+++ b/rust/sdk/src/client.rs
@@ -1096,7 +1096,13 @@ impl<'a> FusedFuture for Subscribe<'a> {
     }
 }
 
-/// Request builder for subscriptions.
+/// Request builder for monotonic subscriptions.
+///
+/// Monotonic subscriptions keep track of the highest sort order
+/// ([`LamportTimestamp`](crate::timestamp::LamportTimestamp) and
+/// [`StreamId`](crate::scalars::StreamId)) seen so far, ending the stream with a
+/// [`SubscribeMonotonicResponse::TimeTravel`](SubscribeMonotonicResponse::TimeTravel)
+/// message if the next event would be out of order.
 ///
 /// Warning: [`SubscribeMonotonic`] implements the [`Future`] trait, as such it can be polled.
 /// Calling _any_ [`SubscribeMonotonic`] function after polling will result in a panic!
@@ -1116,7 +1122,7 @@ impl<'a> SubscribeMonotonic<'a> {
             request: SubscribeMonotonicRequest {
                 query: query.into(),
                 session: SessionId::from("me"),
-                from: OffsetMap::empty(),
+                lower_bound: OffsetMap::empty(),
             },
         }
     }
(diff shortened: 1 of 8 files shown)


rust/sdk/src/client.rs
```diff
@@ -1131,16 +1131,102 @@ impl<'a> SubscribeMonotonic<'a> {
         panic!("Calling SubscribeMonotonic::with_session_id after polling.")
     }
 
-    // TODO: Figure out how this is different from the lower bound
+    // TODO: there's info missing about the difference between Subscribe and SubscribeMonotonic
```
rkuhn, line 1134: SubscribeMonotonic keeps track of the highest sort order (LamportTimestamp+StreamId) seen so far and ends the stream with a TimeTravel message in case the next delivered event would be older.

rust/sdk/src/service/events.rs
```diff
@@ -402,8 +394,7 @@ pub struct SubscribeMonotonicRequest {
     /// previously interrupted stream. In this case, StartFrom::Offsets is used,
     /// otherwise StartFrom::Snapshot indicates that the PondService shall figure
     /// out where best to start out from, possibly sending a `State` message first.
-    #[serde(flatten)]
-    pub from: StartFrom,
+    pub from: OffsetMap,
```
rkuhn, line 397: please add a test that contains the previously written serialisation format and verifies that the new serialisation format is identical

rkuhn (commented):
The “offsets” message is defined to mark the transition from past to future, so it must always contain the OffsetMap of the present.

rust/sdk/src/service/events.rs
jmg-duarte: There actually is a test that caught my error:
https://github.com/Actyx/Actyx/actions/runs/6801666115/job/18493040522?pr=583#step:3:485

rust/sdk/tests/serialization_roundtrips.rs
```diff
@@ -1,24 +1,26 @@
...
 
-fn roundtrip<T: serde::Serialize + serde::de::DeserializeOwned + std::fmt::Debug>(json: Value) -> anyhow::Result<()> {
-    let value: T = from_value(json.clone())?;
-    let serialized = to_value(value)?;
-    anyhow::ensure!(json == serialized, "\nleft:  {:?}\nright: {:?}", json, serialized);
-    Ok(())
+fn roundtrip<T: serde::Serialize + serde::de::DeserializeOwned + std::fmt::Debug>(json: Value) {
+    let value: T = from_value(json.clone()).unwrap();
+    let serialized = to_value(value).unwrap();
+    assert_eq!(json, serialized)
```
rkuhn, line 7: :+1:

rkuhn (approved):

PR #591, opened 2023-11-14
reuse crate::util::setup_logger

Kelerchian:


Kelerchian:
> interesting — I’d like to understand why this makes the difference it makes

The defaults are simply different. But on a second thought, I should review whether the default of crate::util::setup_logger makes sense

Kelerchian:
Ah, so it seems that `crate::util::setup_logger_with_level` has a silly bug. It defaults to "warning" instead of "warn", which is an invalid Directive for tracing's LevelFilter.

rust/actyx/ax/src/libp2p_streaming_response/tests/mod.rs
```diff
@@ -272,6 +259,7 @@ fn err_response_size() {
 
 #[test]
 fn err_request_size() {
+    crate::util::setup_logger();
```
rkuhn, line 262: I added this to try out the logging: with the TestWriter, logging output (including panics, which we feed via `log_panic` into `log`, from which `tracing_log` takes it into `tracing`) will be properly captured by `libtest` and shown as the failure output of a test.

The original purpose of setting up a tracing Subscriber in tests is to enable debugging what does wrong in case of a test failure by using

    RUST_LOG=debug cargo test <some-test-name-prefix>

rust/actyx/ax/src/util/mod.rs
```diff
@@ -34,26 +34,47 @@ use std::{
...
+/// Panics are pushed into the `log` crate. We install the `tracing_log::LogTracer`
+/// so that all `log` logs (including the panics) get forwarded to `tracing`.
+///
+/// Because of the above, it is imperative that we enable logs to be captured by
+/// the `libtest` crate. This is done by calling `with_test_writer` (conditional
+/// on being in the `test` configuration).
+///
+/// To summarise: every test that may need to be debugged using tracing logs should
+/// call this function, which has been designed for this purpose. In particular, it
+/// can be called multiple times (like within the same test binary).
```
rkuhn, line 47: @jmg-duarte FYI

PR #592, opened 2023-11-14
Remove the panics from the scopes

jmg-duarte:


rust/actyx/ax/src/api/events/service.rs
```diff
@@ -637,6 +637,9 @@ mod tests {
         time::timeout,
     };
 
+    /// "Global" timeout for tests in this module
+    const TIMEOUT: u64 = 5;
```
rkuhn, line 641: please use `Duration::from_secs(5)` to make the unit clear

rkuhn (commented):
yes, nice cleanup! just a few small things

web/developer.actyx.com/bin/copy-schemas.js
```diff
@@ -2,7 +2,7 @@ var copydir = require('copy-dir')
 var mkdirp = require('mkdirp')
 
 mkdirp('static/schemas/').then(() => {
-  copydir('../../protocols/json-schema', 'static/schemas/', {}, (err) => {
+  copydir('../../rust/actyx/ax/resources/json-schema', 'static/schemas/', {}, (err) => {
```
rkuhn, line 5: thanks for catching this!

PR #609, opened 2023-11-23
Fix the CI build

jmg-duarte:
- Removes the ARM v6 architecture
- Removes OpenSSL as a dependency and where really needed, replaced it with `rustls-tls`
- Replaces a, so far, uncompiled path from the `ax` binary with a `panic`
- Adapts the Wix, Makefile and friends to build properly

Finally, artifacts (mainly Windows) still need testing


rust/actyx/ax/src/main.rs
```diff
@@ -144,3 +149,75 @@ async fn with_logger<T>(fut: impl Future<Output = T>, verbosity: u8) -> T {
...
+        None => false,
+    };
+    let log_as_json = match log_json {
+        Some(Color::On) => true,
+        Some(Color::Off) => false,
+        Some(Color::Auto) => is_no_tty,
+        None => false,
+    };
+
+    // NOTE: maybe it's time we remove this this
```
rkuhn, line 180: yes, please

wix/actyx-installer/Dockerfile
```diff
@@ -1,3 +1,6 @@
+# This image uses WiX 3.8
+# Check the following link for more information:
+# https://hub.docker.com/r/justmoon/wix/dockerfile
 FROM justmoon/wix
```
rkuhn, line 4: I’d feel better if we depended on the specific hash

rkuhn (approved):
very nice!

wix/actyx-installer/Dockerfile
jmg-duarte: The image hasn't been updated in something like 8 or 4 years. But I get you

PR #610, opened 2023-11-24
Create versioning ADR

jmg-duarte:


dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,106 @@
...
+  - the previously existing `ax` CLI tool,
+    which handled several aspects of running an Actyx swarm;
+  - the `actyx` binary, which was the system that ran "as a node" in the swarm,
+    processing events, handling queries, etc;
+  - the `cert` tool, which was used for licensing purposes and complementing an Actyx
+    swarm's authorization and authentication capabilities.
+- `ax-core` - the library resulting of extracting the core functionality from the
+  `actyx` binary such that it can be exposed for other uses, such as being embedded
+  in an application
+- `node` - the node component from `actyx`, which can be thought of as the inner Actyx
```
rkuhn, line 20: I’m not sold on using “node version”, since almost every programmer will think this refers to node.js. It also focuses on the swarm networking part, while the other interesting half is the event API and query language.

Thinking about options I only come up with generic terms like “service version”. We could also make it distinct and specific by calling it “AQL version”.

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,106 @@
+# Versioning Ax and friends
```
rkuhn, line 1: please add author and date at the top

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,106 @@
...
+Versions will keep being updated by the `release` tool, which will require updating
+according to this change.
+
+The `node` version will be kept in a Rust file to be managed by the `release` tool.
+
+For example:
+
+```rust
+pub const NODE_VERSION: &str = "2.17.0"
+```
```
rkuhn, line 106: we might want to add a check to `ax/build.rs` that bails out if it finds that the `ax_core` dependency specifies a different NODE_VERSION than the CARGO_PKG_VERSION (modulo patches)

rkuhn (commented):
laid out nicely!

dev-docs/2023-11-24 versioning ax.md
Kelerchian: Isn't this also encompassing aspects other than AQL that might change?
e.g. how node-to-node communicate, data retention behavior, etc?

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,115 @@
...
+- Features
+  - will create a minor bump in `node`
+  - will create a minor bump in `ax-core`
+  - will create a minor bump in `ax`
+
+- Fixes
+  - will create a patch bump to `ax-core`
+  - will create a patch bump to `ax` (from the dependency version update)
+
+`ax`
```
Kelerchian, line 97: What about breaking changes in `ax` and `node`?

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,115 @@
...
+
+- `ax-core` at version `0.2.0`
+- `node` at version `2.17`
+- `ax` at version `2.17.0`
+
+From the previous list, the new publishing "rules" become:
+
+`ax-core`
+- Breaking changes
+  - will create a major bump in `ax-core`
```
Kelerchian, line 81: As far as I understand (unless I'm wrong), will a major bump in `ax-core` immediately result in a bump in `ax` too?
Or, despite being a "monorepo" setup, will it be possible to defer `ax` bump due to cargo's capability to use the previous version of `ax-core` in `ax`?

jmg-duarte: It's possible to defer

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,126 @@
...
+    which handled several aspects of running an Actyx swarm;
+  - the `actyx` binary, which was the system that ran "as a node" in the swarm,
+    processing events, handling queries, etc;
+  - the `cert` tool, which was used for licensing purposes and complementing an Actyx
+    swarm's authorization and authentication capabilities.
+- `ax-core` - the library resulting of extracting the core functionality from the
+  `actyx` binary such that it can be exposed for other uses, such as being embedded
+  in an application
+- `databank` - the old `node` component from `actyx`, which can be thought of as the inner Actyx
+  "stack". We work off of the following definition:
```
rkuhn, line 26:
```suggestion
  "stack", comprising data storage, network protocols, querying facilities etc. We work off of the following definition:
```

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,126 @@
...
+around the old `actyx` version and to keep everything working, we absolutely need to
+keep that version number, question is where and how.
+
+If we were going to do "self-publishing" only — i.e. as we were doing with Actions +
+Actyx Releases — we could get away with using clap/structopt's features to distinguish
+the versions of commands, but we are going to be publishing `ax` on crates.io, and that
+warrants more care when publishing, since it means that users will be consuming this
+by means of `cargo install ax`, and we need to be careful not to break their use cases.
+
+Furthermore, the core "runnable" functionality — i.e. the `node` — was part of `actyx`
```
rkuhn, line 59:
```suggestion
Furthermore, the core "runnable" functionality — i.e. the `node` / `databank` — was part of `actyx`
```

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,126 @@
...
+but is now part of `ax-core` which needs to be versioned as a library, as such, changes
```
rkuhn, line 60:
```suggestion
but is now part of `ax-core` which needs to be versioned as a library based on its Rust function signatures. As such, changes
```

dev-docs/2023-11-24 versioning ax.md
```diff
@@ -0,0 +1,126 @@
...
+## The Solution
+
+To that end, we have decided to publish `ax-core` with its own version, starting at
+`0.2.0` and inside it, the `databank` will keep the existing `actyx` version, the CLI
+will move with the `databank` version.
+
+Following this, we now have:
+
+- `ax-core` at version `0.2.0`
+- `databank` at version `2.17`
```
rkuhn, line 72: I think this should be `2.17.0` as per the statement below.

PR #612, opened 2023-11-27
Implement the versioning ADR

jmg-duarte:


rust/release/src/publisher.rs
```diff
@@ -442,6 +442,66 @@ fn mk_blob_tuples(release: &Release, hash: &Oid, os_arch: OsArch) -> Vec<(Source
         (Product::TsSdk, _) => {}
         (Product::RustSdk, _) => {}
         (Product::CSharpSdk, _) => {}
+        (Product::AxCore, _) => {}
+        (Product::Ax, OS::android) => {
+            out.push((
+                format!("{}-binaries/actyx.apk", os_arch.os,),
```
Kelerchian, line 448: It just crossed my mind the question: With android what does `ax`-like app do?

jmg-duarte: A "true" `ax`: panic
The `actyx.apk` is a bundled ax-core pretty much

rust/actyx/ax-core/build.rs
```diff
@@ -1,3 +1,26 @@
+use std::process::Command;
+
+fn git_hash() -> String {
+    // https://stackoverflow.com/a/5737794/576180
+    let out = Command::new("git")
+        .arg("status")
+        .arg("--porcelain")
+        .output()
+        .expect("Error running git status --porcelain");
```
rkuhn, line 9: We can’t expect this for everyone running `cargo install ax`, and it would be meaningless in that context anyway. I’d suggest injecting this environment variable in github actions, not from build.rs.

rust/actyx/ax/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 edition = "2021"
 name = "ax"
-version = "0.1.0"
+version = "2.17.0"
```
rkuhn, line 4: please add a build.rs check that this string matches `ax_core::DATABANK_VERSION`

rust/actyx/ax-core/src/node/mod.rs
```diff
@@ -51,6 +51,8 @@ use std::{
 use structopt::StructOpt;
 use util::init_panic_hook;
 
+pub const DATABANK_VERSION: &str = "2.17.0";
```
rkuhn, line 54: this constant should be exported at the crate top level

rust/release/src/os_arch.rs
```diff
@@ -127,6 +127,24 @@ impl OsArch {
         Arch::current() == arch
     }
 
+    /// As `ax` was introduced, the `linux-arm` architecture was dropped.
+    /// This method is in every way the same as `all`,
+    /// except it does not return said architecture.
```
rkuhn, line 132: This description makes sense (don’t yet know why we still need `linux-arm` in `all`), so I’d have the implementation filter the result of `Self::all()`.

rust/release/src/versions_file.rs
```diff
@@ -205,7 +205,11 @@ impl VersionsFile {
             let product = current.release.product;
             let repo = Repository::open_from_env()?;
             let ts = repo.find_commit(current.commit)?.time().seconds();
-            let time = chrono::Utc.timestamp(ts, 0).to_rfc3339_opts(SecondsFormat::Secs, true);
+            let time = chrono::Utc
+                .timestamp_opt(ts, 0)
+                .single()
+                .expect("a single timestamp")
```
rkuhn, line 211: my understanding is that this string should give the reason, not the requirement (i.e. “git timestamps are valid”)

rkuhn (commented):
This still lacks the code to write the `ax` and `ax-core` versions from `versions` into the respective source files, right?

rust/actyx/ax-core/build.rs
jmg-duarte: And on missing we keep it empty or maybe tag the build like "cargo_install"?

rust/release/src/os_arch.rs
jmg-duarte: Maybe I can remove it, to be honest, I just did the smallest possible change that _definitely_ doesn't break anything.

rust/release/src/versions_file.rs
jmg-duarte: Your suggestion is definitely better, but it's also true that it expected a single timestamp 😅

rust/actyx/ax-core/build.rs
rkuhn: `ax 2.18.0-cargo_install-linux-x86_64-release` looks good to me

rust/actyx/ax-core/src/util/version.rs
```diff
@@ -48,15 +48,42 @@ pub struct NodeVersion {
...
-    None => "",
+    // This is for cargo installations and builds
+    None => "cargo",
 };
 
 lazy_static! {
     pub static ref VERSION: String = NodeVersion::get().to_string();
 }
 
+const fn os() -> &'static str {
```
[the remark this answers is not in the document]
jmg-duarte: Nice catch!

rust/actyx/ax-core/src/util/version.rs
```diff
@@ -48,15 +48,21 @@ pub struct NodeVersion {
...
+    // This is for cargo installations and builds
+    None => "cargo",
 };
 
 lazy_static! {
     pub static ref VERSION: String = NodeVersion::get().to_string();
 }
 
+pub const ARCH: &str = env!("TARGET_ARCH");
+pub const OS: &str = env!("TARGET_SYS");
```
rkuhn, line 64: This labels macos as `darwin`, which previously was called `macos`. I think this can be fixed with a `match`.

rust/actyx/ax-core/src/util/version.rs
```diff
@@ -48,15 +48,21 @@ pub struct NodeVersion {
...
+// https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+const GIT_HASH: &str = match option_env!("GIT_HASH") {
     Some(hash) => hash,
-    None => "",
+    // This is for cargo installations and builds
+    None => "cargo",
 };
 
 lazy_static! {
     pub static ref VERSION: String = NodeVersion::get().to_string();
```
rkuhn, line 60: I think `struct NodeVersion` can be removed and this can be made into `const VERSION` using `concat!()`.

rust/actyx/ax/tests/version_compat.rs
```diff
@@ -165,14 +164,14 @@ fn setup() -> &'static Binaries {
...
-        Arch::x86_64 => "amd64",
-        Arch::aarch64 => "arm64",
-        Arch::arm => "arm",
-        Arch::armv7 => "armhf",
-        x => panic!("unsupported arch: {}", x),
-    };
-    let name = format!("{}-{}-linux-{}", package, version, arch);
+    // This is a bit of an abuse but we're currently running this only on 64-bit Linux
+    // hence taking the liberty to "force" this here. For a reference, the previous
+    // mapping was:
```
rkuhn, line 169: I’d prefer keeping the mapping in here, match on the `ARCH` using string literals

rust/actyx/swarm/harness/src/bin/versions.rs
```diff
@@ -50,14 +50,7 @@ mod versions {
...
             } else {
-                let arch = match Arch::current() {
-                    Arch::x86_64 => "amd64",
-                    Arch::aarch64 => "arm64",
-                    Arch::arm => "arm",
-                    Arch::armv7 => "armhf",
-                    x => panic!("unsupported arch: {}", x),
-                };
-                let name = format!("actyx-{}-linux-{}", version, arch);
+                let name = format!("actyx-{}-linux-{}", version, ARCH);
```
rkuhn, line 53: This needs the same mapping as above; I think this should be provided in `version.rs`.

rust/actyx/ax-core/build.rs
```diff
@@ -1,35 +1,15 @@
...
-        .arg("rev-parse")
-        .arg("HEAD")
-        .output()
-        .expect("Error running git rev-parse --short HEAD");
-
-    let hash = String::from_utf8_lossy(&git_head_out.stdout).trim().to_string();
-
-    format!("{}{}", hash, dirty)
-}
+use std::vec::Vec;
```
rkuhn, line 1: this isn’t needed, right?

rust/actyx/ax-core/build.rs
```diff
@@ -1,35 +1,15 @@
...
 
+// We need this build script to propagate the target variables through compilation
+// instead of relying on #[cfg(target_arch = "...")] which isn't always "guessable"
+// this way, we don't guess, we use what Cargo uses, we split and propagate.
+// See the following link for more information:
+// https://doc.rust-lang.org/cargo/appendix/glossary.html#target
```
rkuhn, line 7: To me this doesn’t explain the issue in full: What would be wrong with using `#[cfg(target_os = ... )]` and `#[cfg(target_arch = ...)]` in `version.rs` at this point? The previous need for AX_ARCH stemmed only from wanting to distinguish arm6 from arm7, but we don’t need that anymore. This means `ax_core` wouldn’t have build.rs at all, which I consider a good thing.

rust/actyx/ax-core/build.rs
jmg-duarte: The import or the vec overall?

rust/actyx/ax/tests/version_compat.rs
jmg-duarte: The mapping is misleading in our process because it doesn't do anything as I explained in the comment. I'll bring it back though

rust/actyx/swarm/harness/src/bin/versions.rs
jmg-duarte: Yeah, I'm still finishing this, I need to push from time to time so I get the code on the Hetzner boz

rust/actyx/ax-core/src/util/version.rs
jmg-duarte: Most likely, yes

rust/actyx/ax-core/build.rs
jmg-duarte: > This means ax_core wouldn’t have build.rs at all, which I consider a good thing.

I agree that it would be good, but it isn't that simple.

1. The issue with using `#[cfg(target_arch = ...)]` is that it is error prone, it was causing issues in the Android build because the arch wasn't in the list (was falling through to the panic) and I wasn't able to simply check which one it was.
2. The `build.rs` is required because the environment variables from cargo (like `TARGET`) are available during building, _not compilation_. They need to be propagated to `rustc` so they're available later for the versioning string.

rust/release/src/main.rs
```diff
@@ -44,16 +45,16 @@ struct Opts {
     cmd: Command,
 }
 #[derive(Parser)]
-#[clap(version = "1.0", author = "Actyx AG", about = "Releases from Cosmos")]
+#[clap(version = "1.0", author = env!("CARGO_PKG_AUTHORS"), about = env!("CARGO_PKG_DESCRIPTION"))]
```
rkuhn, line 48: can also use `, author, about` with the `cargo` feature of `clap`

rust/release/src/main.rs
```diff
@@ -300,6 +304,28 @@ Overview:"#
...
+                            eprint!("0.1) Writing new version to \"{}\" ... ", ax_cargo.display());
+                            update_package_version(&ax_cargo, &new_version)?;
+                            repo.add_file(&ax_cargo)?;
+                        }
+                        Product::AxCore => {
+                            eprint!("0.2) Writing new version to \"{}\" ... ", ax_core_cargo.display());
+                            update_package_version(&ax_core_cargo, &new_version)?;
+                            repo.add_file(&ax_core_cargo)?;
+                        }
+                        // We're not updating TOMLs for anything else
```
rkuhn, line 324: What about updating DATABANK_VERSION? If they aren’t updated at the same time, the build will fail.

Here I suggest again putting that constant into a single-line Rust file so that it can trivially be overwritten.

rust/release/src/main.rs
```diff
@@ -432,3 +458,19 @@ Overview:"#
...
+    // Read the toml
+    let cargo_toml_contents = if !path.is_file() {
+        bail!("{:?} is not a file", path);
+    } else {
+        std::fs::read_to_string(path)?
+    };
+    // Parse it
+    let mut cargo_toml = cargo_toml_contents.parse::<Document>()?;
+    // Update the value
+    cargo_toml["package"]["version"] = toml_edit::value(version.to_string());
```
rkuhn, line 472: nice!

rust/release/src/main.rs
```diff
@@ -432,3 +468,19 @@ Overview:"#
     }
     Ok(())
 }
+
+fn update_package_version(path: &PathBuf, version: &Version) -> Result<(), Error> {
+    // Read the toml
+    let cargo_toml_contents = if !path.is_file() {
+        bail!("{:?} is not a file", path);
```
Kelerchian, line 475: nitpick: this logic should have been covered by `std::fs::read_to_string`

rust/actyx/ax/tests/version_compat.rs
```diff
@@ -185,6 +182,10 @@ fn download(package: &str, bin: &str, version: Version, dst_dir: &Path, may_skip
     }
 
     let resp = reqwest::blocking::get(&url).unwrap_or_else(|e| panic!("making request to {}: {}", url, e));
+    if resp.status() == reqwest::StatusCode::NOT_FOUND {
+        panic!("did not find {}", url);
```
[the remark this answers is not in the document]
jmg-duarte: I did, it panicked during tests for an unpublished binary with a bad error, hence the code.

rust/release/src/main.rs
jmg-duarte: 👍

rust/release/src/main.rs
```diff
@@ -300,6 +304,38 @@ Overview:"#
                 println!("-------------------------\n");
                 println!("Branch to create {}", branch_name);
             } else {
+                // This pathing abuses the fact that we control where this is run from: CI
+                // and in CI this is *usually* run in the rust/release path
+                let ax_cargo = PathBuf::from("../actyx/ax/Cargo.toml").canonicalize()?;
```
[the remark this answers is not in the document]
Kelerchian: Believe it or not, I owe this knowledge to `nodejs`

rust/actyx/ax-core/src/lib.rs
```diff
@@ -14,3 +14,5 @@ pub mod settings;
 pub mod swarm;
 pub mod trees;
 pub mod util;
+
+pub use node::DATABANK_VERSION;
```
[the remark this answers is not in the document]
Kelerchian: Ah, understood. Thanks

rust/release/src/main.rs
```diff
@@ -304,18 +304,26 @@ Overview:"#
...
                 println!("Branch to create {}", branch_name);
             } else {
-                // This pathing abuses the fact that we control where this is run from: CI
-                // and in CI this is *usually* run in the rust/release path
-                let ax_cargo = PathBuf::from("../actyx/ax/Cargo.toml").canonicalize()?;
-                let ax_core_cargo = PathBuf::from("../actyx/ax-core/Cargo.toml").canonicalize()?;
+                // We expect the binary to not have been moved outside of the target folder
+                let cwd = current_exe()?;
+                let actyx_folder = cwd
+                    .ancestors()
```
rkuhn, line 310: TIL: I am my own ancestor …

rust/release/src/main.rs
```diff
@@ -304,18 +304,26 @@ Overview:"#
...
+                    .join(PathBuf::from("actyx/ax-core/Cargo.toml"))
+                    .canonicalize()?;
+
                 for (product, v) in new_versions {
                     let new_version = v.new_version.unwrap();
                     match product {
                         Product::Ax => {
                             eprint!("0.1) Writing new version to \"{}\" ... ", ax_cargo.display());
                             update_package_version(&ax_cargo, &new_version)?;
+                            let version_rs = PathBuf::from("../actyx/ax-core/node/version.rs").canonicalize()?;
```
rkuhn, line 324: please use `actyx_folder` here as well

rust/release/src/main.rs
```diff
@@ -304,18 +304,26 @@ Overview:"#
                 println!("-------------------------\n");
                 println!("Branch to create {}", branch_name);
             } else {
-                // This pathing abuses the fact that we control where this is run from: CI
-                // and in CI this is *usually* run in the rust/release path
-                let ax_cargo = PathBuf::from("../actyx/ax/Cargo.toml").canonicalize()?;
-                let ax_core_cargo = PathBuf::from("../actyx/ax-core/Cargo.toml").canonicalize()?;
+                // We expect the binary to not have been moved outside of the target folder
+                let cwd = current_exe()?;
+                let actyx_folder = cwd
```
rkuhn, line 309: If my counting is correct, then this should be named `rust_folder` instead.

rkuhn (approved):
LGTM, merge after addressing final nitpicks

rust/release/src/main.rs
Kelerchian, line 324: It might be better to also use a path relative from `actyx_folder` (or `rust_folder`) here.

versions
```diff
@@ -1,6 +1,6 @@
 # Last releases of all Actyx products
 # Each line contains <release> <commit-hash>
-# The machine-readable product names are: actyx, node-manager,
+# The machine-readable product names are: ax, ax_core, actyx, node-manager,
```
Kelerchian, line 3: `ax-types`?

jmg-duarte: Doesn't exist in this PR yet and it's precisely the ones we need to release by hand

rust/release/src/main.rs
jmg-duarte: It is. Nice catch

versions
Kelerchian: Ah, nvm. I was confused about the purpose of this thing

PR #614, opened 2023-11-29
clap change

Kelerchian:


rust/actyx/ax/src/main.rs
```diff
@@ -11,99 +11,101 @@ use ax_core::{
...
-#[structopt(
+#[derive(clap::Parser, Clone, Debug)]
+#[command(
     name = "ax",
     about = concat!(
         "\nThe ax CLI is a unified tool to manage your ax nodes.\n\n",
         include_str!("../NOTICE")),
     version = ax_core::util::version::VERSION.as_str(),
+    disable_help_subcommand = true,
+    action = ArgAction::Version
```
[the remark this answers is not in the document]
Kelerchian: It prepends the `help` string, including if `help` is not explicitly called.
It is here to retain the old behavior

rust/actyx/ax/tests/cli.rs
```diff
@@ -109,82 +107,32 @@ fn cli_fail_on_missing_identity() {
...
-                .stderr(starts_with(&*first_line(args.join("-"))));
-        }
-        let name = args.join("-");
-        cli()
-            .args(&*args.into_iter().chain(once("--help")).collect::<Vec<_>>())
-            .env("HERE_BE_DRAGONS", "zøg")
-            .assert()
-            .success()
-            .stdout(starts_with(&*first_line(name)));
-    }
```
rkuhn, line 189: Since we want to guarantee semver also for the command-line behaviour, I think we should keep this test (and update the list of sub-commands to test).

rkuhn (commented):
LGTM, but please merge my `rku/clap-4-tweaks` branch again, there’s one more commit from yesterday, and also restore the cli sub-command structure test.

rust/actyx/ax/tests/cli.rs
rkuhn: My suggestion is to use `ax <subcommands>... --version`, assuming that that fails for non-existing subcommands.

Kelerchian: Ok, let me try a bit if that can be used for it. We can also compare them with default help if it makes sense.

Kelerchian: fixed. could you do a quick check @rkuhn?

rkuhn (approved):
yes, great work!

PR #615, opened 2023-11-30
Move `cmd` out of `ax-core`

jmg-duarte:


rust/actyx/ax-core/src/private_key.rs
```diff
@@ -136,6 +157,54 @@ pub fn load_dev_cert(path: Option<PathBuf>) -> ActyxOSResult<DeveloperCertificat
...
+}
+
+impl fmt::Display for KeyPathWrapper {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0.display())
+    }
+}
+
+// NOTE(duarte): there has to be a better way of doing this
+impl TryFrom<&Option<KeyPathWrapper>> for AxPrivateKey {
```
Kelerchian, line 179: I'd rather use a custom function for this instead of `TryFrom`, since the implementation detail involves a side effect with a particular path default_user_identity_path.

rust/actyx/ax-core/src/private_key.rs
```diff
@@ -10,12 +10,33 @@ use crate::{
...
+}
+
+/// Returns the data directory for Actyx. Does not create the folders!
+/// https://docs.rs/dirs/3.0.1/dirs/fn.config_dir.html
+///
+/// Platform    Value                               Example
+/// Linux       $XDG_CONFIG_HOME or $HOME/.config   /home/alice/.config
+/// macOS       $HOME/Library/Application Support   /Users/Alice/Library/Application Support
+/// Windows     {FOLDERID_RoamingAppData}           C:\Users\Alice\AppData\Roaming
+fn get_data_dir() -> ActyxOSResult<PathBuf> {
```
Kelerchian, line 35: Probably better in `util`, but overall,
I think `ax-core` shouldn't care about paths at all, and this better be put in `ax`, consequently default loading mechanism shouldn't be written inside `ax-core`, instead it should be supplied by `ax` e.g. via dependency-injection

Kelerchian (approved):
looks fine + some remarks

rust/actyx/ax-core/src/private_key.rs
jmg-duarte: I agree that this implementation is not ideal but this is quite pervasive in the code base and this is already a big change 🤔

rust/actyx/ax-core/src/private_key.rs
```diff
@@ -136,6 +157,54 @@ pub fn load_dev_cert(path: Option<PathBuf>) -> ActyxOSResult<DeveloperCertificat
...
+
+// NOTE(duarte): there has to be a better way of doing this
+impl TryFrom<&Option<KeyPathWrapper>> for AxPrivateKey {
+    type Error = ActyxOSError;
+    fn try_from(k: &Option<KeyPathWrapper>) -> Result<Self, Self::Error> {
+        if let Some(path) = k {
+            path.0
+                .to_str()
+                .and_then(|s| s.parse::<AxPrivateKey>().ok())
+                .ok_or(ActyxOSError::internal(""))
```
rkuhn, line 186: not sure I recall where this is used, but it feels suspicious that we erase the error message and replace it by a meaningless one

rust/actyx/ax/tests/version_compat.rs
```diff
@@ -29,6 +27,27 @@ use std::{
 use tar::Archive;
 use tempfile::tempdir;
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged)]
+#[allow(non_camel_case_types)]
+#[allow(clippy::upper_case_acronyms)]
+pub enum ActyxCliResult<T> {
```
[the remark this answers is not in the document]
jmg-duarte: From what I tried and found, it isn't. To test internal components through code we need to expose `ax` as a library. Once published, it will be available for anyone to use but we can just "not document it" or tell people "you shouldn't be using me".

rust/actyx/ax-core/src/authority.rs
```diff
@@ -0,0 +1,44 @@
+use std::str::FromStr;
+
+use crate::util::formats::{ActyxOSCode, ActyxOSError, ActyxOSResult};
+use libp2p::{multiaddr::Protocol, Multiaddr};
+use std::net::ToSocketAddrs;
+
+#[derive(Debug, Clone)]
+pub struct Authority {
```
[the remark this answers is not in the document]
jmg-duarte: That might be the case but I lack the knowledge to reply. We can review this together though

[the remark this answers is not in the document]
rkuhn: Okay, reading the code more carefully we _might_ unify it, but it doesn’t really fit very well: `Authority` describes how to reach an admin port, including multiaddr usage (e.g. for relaying), while `PortOrHostPort` describes a specific (set of) TCP/IP SocketAddr to bind a service to (i.e. multiaddr isn’t relevant).

So I’d say we leave it as is.

rkuhn (approved):

rust/actyx/ax-core/src/private_key.rs
jmg-duarte: I forgot to reply. I don't think addressing this right now is pertinent, though I agree. Both this and the previous issue you outlined are relevant but a bit outside the scope of this PR. Should we open an issue?

Kelerchian: https://github.com/Actyx/Actyx/issues/618
^ Absolutely. There's no need to fix this in this PR I think

PR #616, opened 2023-12-05
Extract ax-aql and ax-types from sdk

jmg-duarte:


rust/actyx/ax-aql/src/language/parser/mod.rs
```diff
@@ -664,12 +665,26 @@ impl FromStr for SimpleExpr {
     }
 }
 
-impl FromStr for Timestamp {
+pub struct AqlTimestamp(u64);
```
rkuhn, line 668: slightly inconvenient, but this seems to be a hard limitation of the Rust trait system

rust/actyx/ax-core/src/api/events/service.rs
```diff
@@ -607,7 +607,10 @@ async fn store_line(store: &BanyanStore, line: &str) -> anyhow::Result<()> {
     let line: Line = serde_json::from_str(line)?;
     let timestamp = line
         .timestamp
-        .or_else(|| line.time.and_then(|t| t.parse().ok()))
+        .or_else(|| {
+            line.time
+                .and_then(|t| t.parse::<AqlTimestamp>().map(Timestamp::from).ok())
```
rkuhn, line 612: we could also add `Timestamp::parse()` method (no trait); this would be a great place where to document why there is no `impl FromStr for Timestamp`

rust/actyx/ax-macros/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "ax_macros"
+version = "0.1.0"
+authors = ["Actyx AG <developer@actyx.io>"]
+edition = "2021"
+description = "Tools for interacting with the services of an Actyx node"
```
rkuhn, line 6: “supporting macros for the SDK”?

rust/actyx/ax-types/Cargo.toml
```diff
@@ -0,0 +1,65 @@
+[package]
+name = "ax_types"
+version = "0.1.0"
+authors = ["Actyx AG <developer@actyx.io>"]
+edition = "2021"
+description = "Actyx's core types"
```
rkuhn, line 6: we might want to rename this also: “core types used by and with ax”

rust/actyx/ax-types/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+readme = "README.md"
+documentation = "https://docs.rs/actyx_sdk"
```
rkuhn, line 8: these links need to be renamed to ax as well

rust/actyx/ax-types/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+homepage = "https://developer.actyx.com/"
+repository = "https://github.com/Actyx/Actyx"
+license = "Apache-2.0"
+keywords = ["distributed", "decentralized", "event-sourcing"]
+categories = ["network-programming"]
+
+[features]
+default = ["arb"]
+arb = ["quickcheck"]
+sqlite = ["rusqlite"]
```
rkuhn, line 18: do we need this somewhere else than in the differential dataflow library? (actyxos_data_flow)

rust/actyx/ax-types/src/lib.rs
```diff
@@ -0,0 +1,38 @@
...
+pub mod types;
+
+pub use app_manifest::AppManifest;
+pub use event::{Event, EventKey, Metadata, Opaque, Payload};
+pub use offset::{Offset, OffsetError, OffsetMap, OffsetOrMin};
+pub use scalars::{AppId, NodeId, StreamId, StreamNr};
+pub use tags::{Tag, TagSet};
+pub use timestamp::{LamportTimestamp, Timestamp};
+
+use derive_more::Display;
```
rkuhn, line 28: I prefer the fully-qualified name inside `derive` because it doesn’t pollute the scope

rust/sdk/Cargo.toml
```diff
@@ -12,78 +12,86 @@ license = "Apache-2.0"
...
-[package.metadata.docs.rs]
-features = ["client"]
-
 [features]
-default = ["client"]
-client = ["reqwest", "url", "bytes", "rand"]
-sqlite = ["rusqlite"]
-arb = ["quickcheck", "rand"]
-with-tokio = ["tokio"]
+with_tokio = ["tokio", "tokio-util", "rand"]
```
rkuhn, line 16: since we depend on reqwest, we always depend on tokio, so the feature should be removed

rkuhn (commented):
nice cleanup! the rust/sdk_macros folder can also be removed, right?

rust/sdk/Cargo.toml
```diff
@@ -38,46 +41,57 @@ serde_cbor = "0.11.2"
 url = "2.5.0"
 tracing = "0.1.40"
 
-# This feature provides exponential backoff on requests
-tokio = { version = "1.34.0", optional = true }
+# Feature: with_tokio
+tokio = { version = "1.15.0", features = [
```
Kelerchian, line 45: what's the reason for downgrading 1.34.0 to 1.15.0? (that is if I'm reading this correctly)

rust/sdk/Cargo.toml
```diff
@@ -38,46 +41,57 @@ serde_cbor = "0.11.2"
...
+    "macros",
+    "fs",
+], optional = true }
+tokio-util = { version = "0.6.9", features = ["compat"], optional = true }
+rand = { version = "0.8.5", optional = true }
 
+[dev-dependencies]
+# This one is to make examples work
+asynchronous-codec = "0.7.0"
+structopt = "0.3.26"
```
Kelerchian, line 56: new issue for upgrading this to clap?

rust/sdk/Cargo.lock
```diff
@@ -1138,6 +1231,18 @@ version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404"
 dependencies = [
+ "libc",
+ "rand_chacha",
```
Kelerchian, line 1235: curious, this is the same package with the same checksum, how do its dependencies change?

rust/actyx/ax-aql/src/language/parser/mod.rs
jmg-duarte: I found an alternative though, we can parse on the library side since this parsing method isn't actually used here.
This was the way I found to keep this working, but I found a library that does the parsing for us. I'll push a commit so you can see

rust/actyx/ax-types/Cargo.toml
jmg-duarte: Yes, `SqliteIndexStore` requires `ToSql` on `StreamId`

rust/sdk/Cargo.toml
jmg-duarte: I don't know how necessary it is since it's an example

rust/sdk/Cargo.lock
jmg-duarte: Features (maybe)

rkuhn (approved):
Great work!

PR #623, opened 2023-12-14
add stack overflow guard in Context::eval_from

Kelerchian:


rust/actyx/ax-core/src/runtime/eval.rs
```diff
@@ -145,6 +153,18 @@ impl<'a> Context<'a> {
 
     pub fn eval_from<'b, 'c: 'b>(&'b self, expr: &'c TagExpr) -> BoxFuture<'b, anyhow::Result<Cow<'c, TagExpr>>> {
         async move {
+            // Below block sets a threshold for the remaining stack during
+            // heavily recursive AQL evaluation to avoid crashing the whole
+            // databank process.
+            if let Some(x) = stacker::remaining_stack() {
+                if x < EVAL_REMAINING_STACK_THRESHOLD {
+                    return Err(anyhow!(
```
rkuhn, line 161: I’d use `anyhow::bail!` in such cases.

rust/actyx/ax-core/src/runtime/eval.rs
```diff
@@ -22,6 +22,14 @@ use std::{
     sync::Arc,
 };
 
+// eval_from method below can be heavily recursive when AQL
+// with long tag bin-ops chain is supplied.
+//
+// https://github.com/Actyx/Actyx/issues/608
+//
+// see `eval_from` below
+const EVAL_REMAINING_STACK_THRESHOLD: usize = 200000;
```
rkuhn, line 31: This number is huge, but this is because we’re only using it on one particular function call. I’m not saying this needs to be done now, but eventually we should have a remaining stack depth check at the entry of _every_ function call controlled by user input that may be called recursively.

rkuhn (approved):
This is the very first baby step, thanks for pioneering! Making Actyx resource-safe regarding user input will be a huge and ongoing effort in the future.

rust/actyx/ax-core/src/runtime/eval.rs
Kelerchian: is there any difference between them other than one is a macro-shortcut for the other?

rust/actyx/ax-core/src/runtime/eval.rs
Kelerchian: It is indeed huge. 
Each recursion of eval_from takes 4800-ish.
(and in this tag case, the stack overflow happens on the very first recursive call (among the 1 other in the same branch and 2 others at a different branch)
But I notice that `println!` stops at 170_000, there might be something going on or something unaccounted for.

rust/actyx/ax-core/src/runtime/eval.rs
rkuhn: no difference, it is just shorter, so easier to read

rust/actyx/ax-core/src/runtime/eval.rs
rkuhn: My guess is that it takes ca. 170kB stack space to run one invocation (with all its contained method calls, some of which do not call `eval_from` recursively). If you place such a check into all methods possibly invoked (also indirectly) by `eval_from`, you’ll find where the stack space is actually used.

rust/actyx/ax-core/src/runtime/eval.rs
Kelerchian: ah ok. done

PR #640, opened 2023-12-19
Refactor loading private keys from the CLI

### LINKED ISSUE #620: specific file path-dependency is not obvious in TryFrom impl of AxPrivateKey
I'd rather use a custom function for this instead of `TryFrom`, since the implementation detail involves a side effect with a particular path default_user_identity_path.

_Originally posted by @Kelerchian in https://github.com/Actyx/Actyx/pull/615#discussion_r1412087300_

---

`impl TryFrom<&Option<KeyPathWrapper>> for AxPrivateKey` implies a specific default path as a fallback mechanism. This semantic is unobvious and may lead to a bug or a misuse

jmg-duarte:
fixes #620


Kelerchian (approved):

rust/actyx/ax-core/src/private_key.rs
```diff
@@ -157,54 +176,6 @@ pub fn load_dev_cert(path: Option<PathBuf>) -> ActyxOSResult<DeveloperCertificat
...
-}
-
-// NOTE(duarte): there has to be a better way of doing this
-impl TryFrom<&Option<KeyPathWrapper>> for AxPrivateKey {
-    type Error = ActyxOSError;
-    fn try_from(k: &Option<KeyPathWrapper>) -> Result<Self, Self::Error> {
-        if let Some(path) = k {
-            path.0
-                .to_str()
-                .and_then(|s| s.parse::<AxPrivateKey>().ok())
```
rkuhn, line 185: I think you removed this functionality: I made it so that the key can be passed verbatim on the command line (or in an env var), avoiding the need to store it in a file

Kelerchian: My bad, I did not realize there is this functionality when making the issue. So now I got it why it is named `KeyPathWrapper`

jmg-duarte: Oh I see now. It's super non obvious because the type of the wrapper is a `PathBuf`.
If we are to support both, I prefer having a flag for the inline key and another for the path. The previous way is too overloaded, both in usage and in code.

rkuhn: well, it is somewhat established to read secrets from env vars, though

jmg-duarte: I can re-add the feature with `String` instead of `PathBuf` as the main type. We can then try to parse into an `AxPrivateKey` and if it fails, read it as a path.

Code-wise this would make much more sense, client wise, it's the same

rust/actyx/ax/src/cmd/mod.rs
```diff
@@ -48,15 +47,20 @@ pub struct ConsoleOpt {
     authority: Authority,
     /// File from which the identity (private key) for authentication is read.
```
rkuhn, line 48: ISTR this also stating that the key can be given verbatim — in any case, could you add that? And while you’re at it, could you copy the doc string to all uses below?

rkuhn (approved):
Thanks! Just the doc string needs a little tweak.

rust/actyx/ax/src/cmd/nodes/ls.rs
```diff
@@ -1,23 +1,24 @@
...
+use std::time::Duration;
 
 #[derive(clap::Parser, Clone, Debug)]
 /// show node overview
 pub struct LsOpts {
     /// the IP address or `<host>:<admin port>` of the nodes to list.
     #[arg(name = "NODE", required = true)]
     authority: Vec<Authority>,
-    /// File from which the identity (private key) for authentication is read.
+    /// Authentication identity (private key).
```
Kelerchian, line 17: Maybe, a little reordering
```
File containing identity OR the identity encoded in base64
Identity is a private key used for authentication
Defaults to the path `<OS_CONFIG_FOLDER>/keys/id`
```

rust/actyx/ax/src/cmd/mod.rs
```diff
@@ -46,17 +45,24 @@ pub struct ConsoleOpt {
     /// the IP address or `<host>:<admin port>` of the node to perform the operation on.
     #[arg(name = "NODE", required = true)]
     authority: Authority,
-    /// File from which the identity (private key) for authentication is read.
+    /// Authentication identity (private key).
+    /// Can be base64 encoded or a path to a file containing the key,
+    /// defaults to `<OS_CONFIG_FOLDER>/keys/id`.
```
rkuhn, line 50:
```suggestion
    /// defaults to `<OS_CONFIG_FOLDER>/keys/users/id`.
```

rust/actyx/ax/src/cmd/nodes/ls.rs
rkuhn: I prefer José’s proposal.

PR #648, opened 2023-12-20
Fix the path

jmg-duarte:


rust/release/src/main.rs
```diff
@@ -341,32 +341,34 @@ pub const DATABANK_VERSION: &str = "{}";
...
                             std::process::Command::new("cargo")
                                 .arg("update")
                                 .output()
                                 .expect("failed to execute `cargo update`");
-                            let lockfile = rust_folder
-                                .parent()
-                                .ok_or(anyhow!("failed to get the main Actyx directory"))?
-                                .join("Cargo.lock");
-                            repo.add_file(lockfile)?;
+                            let lockfile = rust_folder.join("Cargo.lock");
```
rkuhn, line 369: isn’t there an `actyx/` missing?

PR #652, opened 2023-12-21
allow skipping on missing files

jmg-duarte:


rust/actyx/ax/tests/version_compat.rs
```diff
@@ -205,7 +205,12 @@ fn download(package: &str, bin: &str, version: Version, dst_dir: &Path, may_skip
 
     let resp = reqwest::blocking::get(&url).unwrap_or_else(|e| panic!("making request to {}: {}", url, e));
     if resp.status() == reqwest::StatusCode::NOT_FOUND {
-        panic!("did not find {}", url);
+        if *may_skip {
+            *may_skip = false;
```
Kelerchian, line 209: a question: why do we have to set may_skip to false here?
Wouldn't it implicitly mean that may_skip can only be applied once for all ax packages detected in the versions file?

jmg-duarte: From my reading of the remaining code, it's precisely that. Only a single release may be missing — the PR one

Kelerchian: Ah, I see. I just re-read and there's one line of comment about it.
I just thought that this could be coded better, but I think this is for another day.

PR #656, opened 2023-12-22
Replace prettytable with comfy_table

jmg-duarte:
Table looks _slightly_ different, not that we can't change it, but letting you guys know
```
┌─────────────────────────────────────────────┬───────────┬───────────────┬────────┬────────┐
│ NODE ID                                     ┆ HOST      ┆ TOPIC         ┆ SIZE   ┆ ACTIVE │
╞═════════════════════════════════════════════╪═══════════╪═══════════════╪════════╪════════╡
│ blDwxEOHbf/NqhNza6aBiG2qCdH.J5q429X9YCbnFTY ┆ localhost ┆ default-topic ┆ 934432 ┆ *      │
└─────────────────────────────────────────────┴───────────┴───────────────┴────────┴────────┘
```

### DIFF
--- a/rust/actyx/ax/src/cmd/mod.rs
+++ b/rust/actyx/ax/src/cmd/mod.rs
@@ -59,20 +59,6 @@ impl ConsoleOpt {
     }
 }
 
-// pub(crate) mod consts {
-//     use prettytable::format::{FormatBuilder, LinePosition, LineSeparator, TableFormat};
-//     lazy_static::lazy_static! {
-//         pub static ref TABLE_FORMAT: TableFormat = FormatBuilder::new()
-//             .column_separator('│')
-//             .borders('│')
-//             .separators(&[LinePosition::Top], LineSeparator::new('─', '┬', '┌', '┐'))
-//             .separators(&[LinePosition::Title], LineSeparator::new('─', '┼', '├', '┤'))
-//             .separators(&[LinePosition::Bottom], LineSeparator::new('─', '┴', '└', '┘'))
-//             .padding(1, 1)
-//             .build();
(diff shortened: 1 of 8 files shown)


tags.py
rkuhn, line 1: sneaky :-)

rkuhn (approved):

rkuhn:
I don’t mind the slightly different look.

tags.py
jmg-duarte: Oops that's just from my bad habit of adding everything (I have aliases, so `gaa` and commit 😅)

rust/actyx/ax/src/cmd/mod.rs
```diff
@@ -59,19 +59,19 @@ impl ConsoleOpt {
...
-            .column_separator('│')
-            .borders('│')
-            .separators(&[LinePosition::Top], LineSeparator::new('─', '┬', '┌', '┐'))
-            .separators(&[LinePosition::Title], LineSeparator::new('─', '┼', '├', '┤'))
-            .separators(&[LinePosition::Bottom], LineSeparator::new('─', '┴', '└', '┘'))
-            .padding(1, 1)
-            .build();
-    }
-}
+// pub(crate) mod consts {
```
Kelerchian, line 62: Is this kept for, in the future, referring back to old table style?

Kelerchian (approved):
No concern on the style change

rust/actyx/ax/src/cmd/mod.rs
jmg-duarte: Nope, because I forgot to delete. Nice catch

PR #666, opened 2024-01-03
Run cargo update when ax' Cargo.toml is updated

jmg-duarte:


rust/release/src/main.rs
```diff
@@ -339,6 +339,17 @@ pub const DATABANK_VERSION: &str = "{}";
                             )?;
                             repo.add_file(&ax_cargo)?;
                             repo.add_file(&version_rs)?;
+
+                            // Update the lockfile
+                            eprintln!("0.2.1) Updating the lockfile");
+                            std::process::Command::new("cargo")
+                                .current_dir(rust_folder.join("actyx/ax"))
```
Kelerchian, line 346: does the difference between `cargo update` inside `actyx` folder and `actyx/ax` folder matter?

Kelerchian (approved):

PR #674, opened 2024-01-10
releasing docs

Kelerchian:


RELEASING.md
```diff
@@ -0,0 +1,36 @@
+# Releasing Guide
+
+AX releasing workflow is semi-guided by the CI logic comprised of [GitHub actions scripts](.github/workflows/), [makefile](./Makefile), and [cosmos-release tool](./rust/release/).
```
rkuhn, line 3: “comprised of” is almost always incorrect: to comprise means to “take together”, i.e. to gather, so “comprised of” would mean “gathered by”.
```suggestion
The AX releasing workflow is semi-guided by the CI logic comprising [GitHub actions scripts](.github/workflows/), [makefile](./Makefile), and [cosmos-release tool](./rust/release/).
```

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+> - Note: Explanations for the text in bold can be found in the [glossary section](#glossary).
+
+1. A release starts when a **changelog commit** is created.
+1. The previous step triggers the creation of a **Release PR**. This PR will initially fail because the version validation detects that a version bump is required.
+1. _[Manually]_ With a developer's assistance, each impacted **product's** version is bumped correctly. These product's version can be found in each of the corresponding cargo.toml or package.json file.
+1. _[Manually]_ the **Release PR** is merged when its build and validation chain is successful.
+1. After **Release PR** is merged to the `master` branch, a publication process will run; it publishes all **products** listed on the previously merged release.
+1. _[Manually]_ A developer publishes `rust` crates that have been changed.
+
+## Glossary:
```
rkuhn, line 16: This should go last in the file.

RELEASING.md
```diff
@@ -0,0 +1,36 @@
+# Releasing Guide
+
+AX releasing workflow is semi-guided by the CI logic comprised of [GitHub actions scripts](.github/workflows/), [makefile](./Makefile), and [cosmos-release tool](./rust/release/).
+
+## General Release Steps
+
+> - Note: Explanations for the text in bold can be found in the [glossary section](#glossary).
+
+1. A release starts when a **changelog commit** is created.
```
rkuhn, line 9: “Created” would probably be understood widely, including when I create the commit on my laptop — but that doesn’t start the release. I’d say it starts when a changelog is pushed to master (which regularly occurs via a PR).

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+1. The previous step triggers the creation of a **Release PR**. This PR will initially fail because the version validation detects that a version bump is required.
+1. _[Manually]_ With a developer's assistance, each impacted **product's** version is bumped correctly. These product's version can be found in each of the corresponding cargo.toml or package.json file.
+1. _[Manually]_ the **Release PR** is merged when its build and validation chain is successful.
+1. After **Release PR** is merged to the `master` branch, a publication process will run; it publishes all **products** listed on the previously merged release.
+1. _[Manually]_ A developer publishes `rust` crates that have been changed.
```
rkuhn, line 14: The process you describe is the old one, you guys changed all this on Monday, right?

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Glossary:
+
+- **`Actyx` Product**: or simply product, refers to publishable artifacts from this repository (`ax`, `ax-core`, `node-manager`, etc).
+- **changelog commit**: a conventional commit with a valid product designation that will trigger a release process and will be included in the release changelog.
```
rkuhn, line 19: This line should describe exactly how a changelog is attached to a commit instead of the consequences this will have down the line. It should probably point to a section below detailing the changelog entry syntax.

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+- **versions file**: is a special file that lists supposedly released versions of **`Actyx` products**. This file is located [here](./versions).
```
rkuhn, line 20: The “supposedly” will be confusing to the reader. Perhaps better remove it and add a remark that released is not the same as published, pointing to the description of the publishing mechanism.

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+- **Release PR**: An automatically generated Pull Request for release purposes. It contains a changelog and the appending of the **version file**.
+- **Databank**: refers to the set of storage rules and the communication protocol inside `ax-core`. Previously this was a standalone binary simply called `Actyx`. A **Databank** has a version number to co-identify whether it has a compatible communication protocol with a peer **Databank**.
```
rkuhn, line 22:
```suggestion
- **databank version**: refers to the set of storage rules, communication protocols, and other behaviours inside `ax-core`.
  This is necessary because the Rust crate version of `ax-core` is dictated by semantic versioning, which regards only the exposed data types and function signatures, i.e. an orthogonal set of concerns.
    The databank version continues to denote what was previously the version of the `Actyx` binary (when `ax` was only its remote management CLI).
```

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Versioning and Development Policy Regarding Version Interdependency
+
+`ax`, `ax-core`, and **Databank** have version interdependency, namely, a major and minor bump in `ax` results in the **Databank** version bump. Because of this, some rules are needed.
```
rkuhn, line 26:
```suggestion
`ax`, `ax-core`, and **databank** have version interdependencies, because of the policy decision that the `ax` version shall match the databank version, where the former has a patch component and the latter has not.
This inspires the following rules:
```

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+
+- **Databank** version reflects the protocol inside while `ax-core` version reflects its API.
```
rkuhn, line 28:
```suggestion
- **databank** version reflects the protocol inside while `ax-core` version reflects its Rust API.
```

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+- **Databank** version inside `ax-core` that is being included in `ax` must be the `ax`'s `major.minor` version format.
```
rkuhn, line 29:
```suggestion
- **databank** version inside `ax-core` that is being included in `ax` must be `ax`'s `major.minor` version part.
```

rkuhn (commented):
Thanks for starting this. I think the structure is good (apart from the placement of the glossary), and it will grow a few more (sub)sections as noted.

The graph still depicts the old release flow and needs to be updated. Here, all actions are triggered by pushes to some branch, right? I’d phrase it as such, the wording “merge to” is less clear.

Kelerchian:
> The graph still depicts the old release flow and needs to be updated. Here, all actions are triggered by pushes to some branch, right?

Not all actions, but the version check happens inside "validate" fires on all updates on PRs (which, when the merging rule is set as it is right now, is an indicator of an upcoming merge).

Although, one thing I think is still missing is the complete protection from:
- Accidentally omitting bump in `ax-core`'s `cargo.toml.version` when the `DATABANK_VERSION` is bumped.
- Correct "next_version" calculation that involves `ax-core`'s and `ax`'s interdependency.
With those, it'll feel much safer to make changes

rkuhn:
Let’s try to complete this soon — we need to get back to releasing. @jmg-duarte could you also review regarding the current state of affairs?

PR #681, opened 2024-02-05
Remove & move some unused/barely used code

jmg-duarte:
During my ax-core reading, I noticed some things that could be moved out to make the core lighter


rkuhn (approved):
LGTM

PR #682, opened 2024-02-28
add parsing of EVENT/WORKFLOW/MACHINE

rkuhn:
This is only the first step, pure syntax. The MACHINE operation yields a TODO variable binding
instead of doing anything useful. Before this becomes usable we must add some form of name checking
at least, ensuring that referenced workflows and event types actually exist. Ideally, we should also
add type-checking of the binder expressions in workflows, because we know the respective event types
now.

- fix snapshot invalidation
- add EVENT type parsing
- add workflow parsing

### DIFF
--- a/rust/actyx/ax-aql/src/language/aql.pest
+++ b/rust/actyx/ax-aql/src/language/aql.pest
@@ -1,4 +1,4 @@
-main_query = { SOI ~ pragma* ~ features? ~ query ~ EOI }
+main_query = { SOI ~ pragma* ~ event* ~ features? ~ query ~ EOI }
 main_tag_expr = { SOI ~ tag_expr ~ EOI }
 main_simple_expr = { SOI ~ simple_expr ~ EOI }
 main_ident = @{ SOI ~ ident ~ EOI }
@@ -7,10 +7,54 @@ main_timestamp = @{ SOI ~ isodate ~ EOI }
 
 // pragmas
 pragma = ${ "PRAGMA" ~ WHITE_SPACE+ ~ feature_word ~ ( !NEWLINE ~ WHITE_SPACE )* ~ ( pragma_multi | pragma_single ) }
-pragma_multi = _{ NEWLINE ~ pragma_multi_text ~ NEWLINE ~ "ENDPRAGMA" ~ NEWLINE }
-pragma_multi_text = ${ ( !( NEWLINE ~ "ENDPRAGMA" ~ NEWLINE ) ~ ANY )* }
+pragma_multi = _{ NEWLINE ~ pragma_multi_text ~ NEWLINE ~ pragma_end ~ NEWLINE }
+pragma_multi_text = ${ ( !( NEWLINE ~ pragma_end ~ NEWLINE ) ~ ANY )* }
+pragma_end = { "ENDPRAGMA" }
 pragma_single = _{ ":=" ~ ( !NEWLINE ~ WHITE_SPACE )* ~ pragma_single_text ~ NEWLINE }
 pragma_single_text = ${ ( !NEWLINE ~ ANY )* }
+
+features = { "FEATURES(" ~ feature_word* ~ ")" }
+feature_word = @{ ( ASCII_ALPHANUMERIC | "ø" )+ }
+
+// types
+type = { type_atom ~ type_post* ~ ( type_infix ~ type_atom ~ type_post* )* }
+type_null = { "NULL" }
+type_bool = { "BOOLEAN" }
+type_number = { "NUMBER" }
+type_timestamp = { "TIMESTAMP" }
+type_string = { "STRING" }
+type_universal = { "UNIVERSAL" }
+type_tuple = { brackl ~ type ~ ( comma ~ type )* ~ brackr }
+type_record = { curlyl ~ ( ( natural | ident | nonempty_string ) ~ colon ~ type ~ comma? )+ ~ curlyr }
+type_paren = { parenl ~ type ~ parenr }
+type_atom = _{ type_null | type_bool | bool | type_number | natural | type_timestamp | type_string | string
+  | type_tuple | type_record | type_paren | type_universal }
+type_array = { "[]" }
+type_dict = { "{}" }
+type_post = _{ type_array | type_dict }
+type_union = { "|" }
+type_intersection = { "&" }
+type_infix = _{ type_union | type_intersection }
+
+// events
+event = { "EVENT" ~ ident ~ event_type ~ tagged? }
+event_type = { type_record } // needed for parser code structure
+tagged = { "TAGGED" ~ tag ~ ( comma ~ tag )* }
+
+// workflows
+workflow = { "WORKFLOW" ~ ident ~ parenl ~ wf_arg ~ ( comma ~ wf_arg )* ~ parenr ~ wf_scope }
+wf_arg = _{ ( role | unique ) ~ ident }
+wf_scope = _{ curlyl ~ wf_step+ ~ curlyr }
+wf_step = _{ wf_event | wf_retry | wf_timeout | wf_parallel | wf_call | wf_compensate | wf_choice }
+wf_event = { ( ident ~ colon )? ~ ( fail | return )? ~ ident ~ at ~ ident ~ wf_binder* }
+wf_binder = { curlyl ~ ident ~ colon ~ ident ~ arrowl ~ ( SELF | simple_expr ) ~ curlyr }
+wf_retry = { retry ~ wf_scope }
+wf_timeout = { timeout ~ duration ~ wf_scope ~ wf_event }
+wf_parallel = { parallel ~ natural? ~ wf_scope }
+wf_call = { match ~ ident ~ parenl ~ ident ~ ( comma ~ ident )* ~ parenr ~ curlyl ~ wf_match* ~ curlyr }
+wf_match = { case ~ ( ident | star ) ~ dblarrowr ~ wf_step+ }
+wf_compensate = { compensate ~ wf_scope ~ with ~ wf_scope }
+wf_choice = { choice ~ curlyl ~ ( case ~ wf_step+ )+ ~ curlyr }
 
 // queries
 query = { "FROM" ~ ( tag_expr ~ query_order? | array ) ~ query_op* ~ "END"? }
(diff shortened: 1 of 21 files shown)


rust/actyx/ax-aql/src/language/parse_utils.rs
```diff
@@ -113,3 +117,77 @@ impl<'a> Ext<'a> for P<'a> {
...
+
+impl<'a, T: PartialEq> PartialEq for Span<'a, T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.value == other.value
+    }
+}
+
+impl<'a, T: Ord> Ord for Span<'a, T> {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        self.value.cmp(&other.value)
```
jmg-duarte, line 150: What is being ordered by this? Shouldn't the `Ord` implementation take into account the span itself rather than just the value?

jmg-duarte (commented):
Some questions and nitpicks.

One thing that is missing is more handwritten tests as they also serve as documentation.

rust/actyx/ax-aql/src/language/aql.pest
```diff
@@ -7,23 +7,68 @@ main_timestamp = @{ SOI ~ isodate ~ EOI }
...
+feature_word = @{ ( ASCII_ALPHANUMERIC | "ø" )+ }
+
+// types
+type = { type_atom ~ type_post* ~ ( type_infix ~ type_atom ~ type_post* )* }
+type_null = { "NULL" }
+type_bool = { "BOOLEAN" }
+type_number = { "NUMBER" }
+type_timestamp = { "TIMESTAMP" }
+type_string = { "STRING" }
+type_universal = { "UNIVERSAL" }
```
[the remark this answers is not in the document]
rkuhn: Yes, this is the ⊤ type, which is called `Any` in Scala, `unknown` in Typescript, `Object` in Java, … Using `ANY` felt wrong because `any` in Typescript is probably most well-known and means something else. And `UNKNOWN` doesn’t convey the right connotations as well, because it implies that knowledge is missing.

rust/actyx/ax-aql/src/language/mod.rs
```diff
@@ -668,6 +715,28 @@ mod for_tests {
...
+            fn String(g: &mut Gen) -> Label {
+                Label::String(Arbitrary::arbitrary(g))
+            }
+            #[allow(non_snake_case)]
+            fn Number(g: &mut Gen) -> Label {
+                Label::Number(Arbitrary::arbitrary(g))
+            }
+            let choices = &[String, Number][..];
+            (g.choose(choices).unwrap())(g)
+        }
```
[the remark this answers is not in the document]
rkuhn: This is a useful pattern: computing `arbitrary(g)` can be very expensive.

rust/actyx/ax-aql/src/language/mod.rs
```diff
@@ -812,6 +921,229 @@ mod for_tests {
...
+            }))
+        }
+    }
+
+    fn ident_chars() -> &'static [char] {
+        static CHOICES: OnceCell<Vec<char>> = OnceCell::new();
+        &CHOICES.get_or_init(|| ('a'..='z').chain('A'..='Z').chain('0'..='9').collect())
+    }
+    const MINUSCULE: Range<usize> = 0..26;
+    const MAJUSCULE: Range<usize> = 26..52;
```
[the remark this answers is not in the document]
rkuhn: Yeah, I know. I still use the proper words wherever I can — call me a language activist.

rust/actyx/ax-aql/src/language/parse_utils.rs
rkuhn: I thought the same, until I saw where it is used: ordering expression trees in the computation of AGGREGATE clauses to find duplicates (and compute only once). This means that including the spans defeats the purpose.

This `Span` is just a container for extra information that isn’t supposed to influence the rest of the program. A more practical problem was the roundtrip test, in which it is … inconvenient to generate the correct spans together with the syntax tree, so that the comparison in the end returns `true`.

rust/actyx/ax-aql/src/language/aql.pest
```diff
@@ -7,23 +7,68 @@ main_timestamp = @{ SOI ~ isodate ~ EOI }
...
+pragma_multi_text = ${ ( !( NEWLINE ~ pragma_end ~ NEWLINE ) ~ ANY )* }
+pragma_end = { "ENDPRAGMA" }
 pragma_single = _{ ":=" ~ ( !NEWLINE ~ WHITE_SPACE )* ~ pragma_single_text ~ NEWLINE }
 pragma_single_text = ${ ( !NEWLINE ~ ANY )* }
 
+features = { "FEATURES(" ~ feature_word* ~ ")" }
+feature_word = @{ ( ASCII_ALPHANUMERIC | "ø" )+ }
+
+// types
+type = { type_atom ~ type_post* ~ ( type_infix ~ type_atom ~ type_post* )* }
```
Kelerchian, line 20: is type_post here for describing something like `number[]`?
In this case, I find it foreign that `type_post` contains `type_dict` which allows `number{}`.
What is this for?

rkuhn: Array is a function from [0, N) ↦ T, tuple does the same but the returned type depends on the index.
Record corresponds to a tuple, but with arbitrary names instead of the first N natural numbers as index.
Dict(ionary) is the corresponding feature to array, but again indexed by name; this is what is called `Record<T, U>` in Typescript, but with T fixed to be `BOOL | NUMBER | STRING`. For symmetry with the array notation, I chose to suffix with the same kind of parentheses used for a record.

```
tuple: [T1, T2, T3]
array: T[]

record: { n1: T1, n2: T2 }
dict:   T{}
```

Kelerchian: Ah I see now

rust/actyx/ax-aql/src/language/aql.pest
```diff
@@ -7,23 +7,68 @@ main_timestamp = @{ SOI ~ isodate ~ EOI }
...
+wf_scope = { curlyl ~ wf_step+ ~ curlyr }
+wf_step = _{ wf_event | wf_retry | wf_timeout | wf_parallel | wf_call | wf_compensate | wf_choice }
+wf_event = { ( wf_state ~ colon )? ~ ( fail | return )? ~ ident ~ at ~ ident ~ wf_binder* }
+wf_state = { ident }
+wf_binder = { curlyl ~ ident ~ colon ~ ident ~ arrowl ~ simple_expr ~ curlyr }
+wf_retry = { retry ~ wf_scope }
+wf_timeout = { timeout ~ duration ~ wf_scope ~ wf_event }
+wf_parallel = { parallel ~ natural? ~ wf_cases }
+wf_call = { match ~ ident ~ parenl ~ ident ~ ( comma ~ ident )* ~ parenr ~ curlyl ~ wf_match* ~ curlyr }
+wf_match = { case ~ ( ident | star ) ~ dblarrowr ~ wf_step+ }
```
Kelerchian, line 56: nitpick for UX reasons: I imagine `wf_step+` here will look ambiguous. it indicates two syntaxes of the atomicity of a block/function body: 1.) being encapsulated inside `{}` and 2.) otherwise.

Switching between the two may cause unintentional cognitive load.

Shouldn't it be a `(wf_step | wf_scope)` instead, or for some simplicity `wf_scope` only?

rust/actyx/ax-aql/src/language/aql.pest
```diff
@@ -7,23 +7,68 @@ main_timestamp = @{ SOI ~ isodate ~ EOI }
...
+event = { "EVENT" ~ ident ~ event_type ~ tagged? }
+event_type = { type_record } // needed for parser code structure
+tagged = { "TAGGED" ~ tag ~ ( comma ~ tag )* }
+
+// workflows
+workflow = { "WORKFLOW" ~ ident ~ parenl ~ wf_arg ~ ( comma ~ wf_arg )* ~ parenr ~ wf_scope }
+wf_arg = { ( role | unique ) ~ ident }
+wf_scope = { curlyl ~ wf_step+ ~ curlyr }
+wf_step = _{ wf_event | wf_retry | wf_timeout | wf_parallel | wf_call | wf_compensate | wf_choice }
+wf_event = { ( wf_state ~ colon )? ~ ( fail | return )? ~ ident ~ at ~ ident ~ wf_binder* }
```
Kelerchian, line 49: where should I see what `some_state: FAIL ident1@ident2` mean for example?

Kelerchian (commented):
some remarks from reading `aql.pest`

rust/actyx/ax-aql/src/language/aql.pest
```diff
@@ -1,4 +1,4 @@
-main_query = { SOI ~ pragma* ~ features? ~ query ~ EOI }
+main_query = { SOI ~ pragma* ~ event* ~ workflow* ~ features? ~ query ~ EOI }
```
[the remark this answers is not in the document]
rkuhn: `features` will be removed as soon as we get around removing it from all clients/libraries

rust/actyx/ax-aql/src/language/aql.pest
rkuhn: I played with it a bit and since each case is terminated either by `CASE` or `}` I didn’t find it ambiguous.

But your questions reminds me: `wf_scope` isn’t actually about scoping, but I only figured that out after I’d started writing the grammar. Should probably be called `wf_block`.

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -0,0 +1,108 @@
...
+    Intersection(Arc<(Type, Type)>),
+    Array(Arc<Type>),
+    Dict(Arc<Type>),
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+pub enum TypeAtom {
+    Null,
+    Bool(Option<bool>),
+    Number(Option<u64>),
```
jmg-duarte, line 26: Doesn't this make all numbers, necessarily a `u64`?

rkuhn: Adding support for proper numbers is still possible, but I defer it because it is a huge effort.
BigInt/Decimal/Float are already supported by cbor-data, but all computation is still missing.

jmg-duarte: So, refinements only work on u64 and the other cases can be floats?

rkuhn: No, currently the grammar and parser only accepts u64, which is a proper subset, so it can be expanded in the future.

rust/actyx/ax-aql/src/language/aql.pest
Kelerchian, line 20: I just realized that there is no `undefined` type which, in TS/JS, allows optional fields.
It is useful when event definition changes in a non-breaking manner so that previously persisted events can still match and work for a particular workflow.
cc: @jmg-duarte

rkuhn: There is NULL — I think `undefined` is a huge mistake and don’t plan on repeating it. Optional fields should be solved differently, for example using properly defined event schema migrations. Or using default values in the type declaration.

PR #683, opened 2024-03-05
Add validator

jmg-duarte:


rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -10,28 +10,28 @@ use once_cell::sync::Lazy;
...
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Hash)]
 pub enum Type {
     Atom(TypeAtom),
     Union(Arc<(Type, Type)>),
     Intersection(Arc<(Type, Type)>),
     Array(Arc<Type>),
     Dict(Arc<Type>),
+    Tuple(NonEmptyVec<Type>),
+    Record(NonEmptyVec<(Label, Type)>),
```
rkuhn, line 21: shouldn’t this require changes in the `impl Arbitrary`?

rust/actyx/ax-core/src/runtime/operation/validation.rs
```diff
@@ -0,0 +1,754 @@
...
+        Err(anyhow::anyhow!("expected a record (dict)"))
+    }
+}
+
+// Using logic rules and De Morgan laws to rewrite intersections and unions before we reach this method
+// would (probably) allow us to write a better (i.e. simpler) and more performant validator
+fn validate_union(value: &CborValue, ty: &(Type, Type)) -> Result<(), anyhow::Error> {
+    // Since the union operation is commutative, we always evaluate the most specific type first
+    // TODO: add tests
+    match ty.borrow() {
```
rkuhn, line 191: why not just try the left one first, then the right one?

rust/actyx/ax-aql/src/language/types.rs
jmg-duarte: Yes, this is still a WIP and that part doesn't compile yet. I need to figure out the whole `arb!` macro

rust/actyx/ax-core/src/runtime/operation/validation.rs
jmg-duarte: I tried that approach first. IMO it becomes very hard to read due to nesting and this one provides an overall structure that changes little over the course of the function

rust/actyx/ax-aql/src/language/workflow.rs
```diff
@@ -235,6 +236,14 @@ fn r_event(p: P<'_>) -> Result<WorkflowStep<'_>, anyhow::Error> {
     let _at = p.next().unwrap();
     let participant = Span::make(p.next().unwrap(), |mut p| Ok(Ident(p.non_empty_string()?)))?;
     let binders = p.map(r_binding).collect::<Result<Vec<_>, _>>()?;
+    for binding in &binders {
+        if !matches!(binding.value(), SimpleExpr::Indexing(_)) {
```
Kelerchian, line 240: shouldn't it also check whether it is `Ind { head, .. }` where head is `SimpleExpr::Var("_")`?

rust/actyx/ax-aql/src/language/workflow.rs
```diff
@@ -201,6 +236,20 @@ fn r_event(p: P<'_>) -> Result<WorkflowStep<'_>, anyhow::Error> {
     let _at = p.next().unwrap();
     let participant = Span::make(p.next().unwrap(), |mut p| Ok(Ident(p.non_empty_string()?)))?;
     let binders = p.map(r_binding).collect::<Result<Vec<_>, _>>()?;
+    for binding in &binders {
+        if let SimpleExpr::Indexing(index) = binding.value() {
+            if let SimpleExpr::Variable(ref var) = *index.head {
+                if var.0 != "_" {
+                    return Err(anyhow!("indexing bindings can only index `_`"));
```
jmg-duarte, line 243:
```
_.a.b
```

rust/actyx/ax-core/src/runtime/operation/validation.rs
```diff
@@ -0,0 +1,653 @@
...
+        } else {
+            Ok(())
+        }
+    } else {
+        Err(anyhow::anyhow!("type mismatch, expected value to be a string"))
+    }
+}
+
+/// Check if a CBOR value is an array. Can also be used to check for tuples (following RFC 7049).
+fn validate_array(value: &CborValue) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
jmg-duarte: I can change to the full path as you showed. Otherwise I find that type aliasing `Result<T, E>` to be an anti pattern. It worsens readability and comprehension to save characters, hence originally avoiding it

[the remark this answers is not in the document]
rkuhn: I agree that `use anyhow::Result` shouldn’t be done, but using the fully qualified type is both explicit and easy to read — arguably easier than `Result<(), anyhow::Error>`.

rust/actyx/ax-aql/src/language/query_impl.rs
```diff
@@ -137,11 +137,10 @@ mod tests {
...
 
-        let start_label = Label::try_from("start").expect("should be a valid label");
-        assert!(map.get(&start_label).is_some());
-
-        let done_label = Label::try_from("done").expect("should be a valid label");
-        assert!(map.get(&done_label).is_none());
+        assert!([("start", true), ("done", false)]
+            .into_iter()
+            .flat_map(|(label, exists)| Label::try_from(label).map(|label| (label, exists)))
+            .all(|(label, exists)| map.contains_key(&label) == exists));
```
rkuhn, line 143: This is quite noisy and yields no information upon failure. I typically write a little helper to make the test clean and expressive.
```rust
let types = query.get_used_event_types().map(|x| x.0).collect::<BTreeSet<_>>();
assert_eq!(types, t(&["start"]));
```
with a definition of `t` that takes a slice and produces a `BTreeSet<Label>`.

rust/actyx/ax-aql/src/language/query_impl.rs
```diff
@@ -137,11 +137,10 @@ mod tests {
...
         assert!(!map.is_empty());
 
-        let start_label = Label::try_from("start").expect("should be a valid label");
-        assert!(map.get(&start_label).is_some());
-
-        let done_label = Label::try_from("done").expect("should be a valid label");
-        assert!(map.get(&done_label).is_none());
+        assert!([("start", true), ("done", false)]
+            .into_iter()
+            .flat_map(|(label, exists)| Label::try_from(label).map(|label| (label, exists)))
```
rkuhn, line 142: This `try_from` must succeed, so `.unwrap()` is the correct choice.

rust/actyx/ax-core/src/runtime/operation/validation.rs
```diff
@@ -95,34 +95,68 @@ fn validate_string(value: &CborValue, string_refinement: &Option<String>) -> Res
...
+            }
+        }
         Ok(())
     } else {
         Err(anyhow::anyhow!("type mismatch, expected value to be an array"))
     }
 }
 
-fn validate_tuple(value: &CborValue, length: usize) -> Result<(), anyhow::Error> {
+fn validate_tuple(value: &CborValue, ty: &NonEmptyVec<Type>) -> anyhow::Result<()> {
```
rkuhn, line 111: since the non-emptiness isn’t used in this function, it would be better to accept `&[Type]`

rust/actyx/ax-core/src/runtime/operation/validation.rs
```diff
@@ -321,16 +351,26 @@ mod test {
             builder.encode_u64(10).encode_u64(100);
         });
         let cbor = cbor.decode();
-        assert!(validate_array(&cbor).is_ok());
+        validate_array(&cbor, &Type::Atom(TypeAtom::Number(None))).unwrap();
+        assert!(validate_array(&cbor, &Type::Atom(TypeAtom::Number(None))).is_ok());
```
rkuhn, line 355: these two do the same; I prefer the first version because upon failure I can immediately see the cause — I very rarely use `assert!()`, and if I do, I never do it with just one argument (I also never trust a test without having made it fail, which also shows me the quality of the failure output)

rkuhn (commented):
Yes, nearly there, just a few test upgrades in case someone ever breaks a test.

rust/actyx/ax-core/src/runtime/operation/validation.rs
jmg-duarte: This was a leftover of a debug session. Ironically, for the reason you meant. I guess it's time I rewrite them 😅

jmg-duarte: I tried both approaches and I find the `assert!` with a second parameter better than the first. 
The first does provide much more output but IMO it doesn't help, having a short message explaining the assumption that was broken and the respective line works out better (in this case).

rust/actyx/ax-core/src/runtime/operation/validation.rs
```diff
@@ -379,15 +420,17 @@ mod test {
...
+        assert!(
+            validate_tuple(
+                &cbor,
+                &[
+                    Type::Atom(TypeAtom::Number(None)),
+                    Type::Atom(TypeAtom::Number(Some(100)))
+                ]
+            )
+            .is_ok(),
+            "TUPLE contents are not (NUMBER, NUMBER(100)"
```
rkuhn, line 432: We probably talked past each other: this message doesn’t help when the test fails — what I need to know in that case is the error message (and possibly what the `&cbor` actually contained). I still think that `.unwrap()` is the best way to get this information.

rkuhn (approved):

rust/actyx/ax-core/src/runtime/operation/validation.rs
rkuhn: Reading this I’m quite sure we’re talking past each other. What I mean is this:
```rust
fn test_validate_tuple() {
    let cbor = CborBuilder::new().encode_array(|builder| {
        builder.encode_u64(10).encode_u64(100);
    });
    let cbor = cbor.decode();
    validate_tuple(&cbor, &[Atom(Number(None)), Atom(Number(Some(100)))]).unwrap();
}
```

PR #684, opened 2024-03-07
temporary QueryWorkflowAnalysis code

Kelerchian:


rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,116 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+
+impl<'a> QueryWorkflowAnalysis<'a> {
+    pub fn from(query: &'a Query) -> Self {
+        Self {
+            query,
+            workflow_tracker: Default::default(),
+        }
+    }
+
+    pub fn check(&'a self, query: &'a Query) -> Vec<QueryWorkflowAnalysisError> {
```
rkuhn, line 748: The `query` argument should be the same as `self.query`, right? In that case it should be removed.

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,116 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+        let mut errors = Vec::new();
+        query
+            .workflows
+            .iter()
+            .for_each(|(_ident, workflow)| errors.extend(self.check_workflow(workflow)));
+        errors
```
rkuhn, line 754: I’d use `self.query.workflows.values().flat_map(|wf| self.check_workflow(wf)).collect()`

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,116 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+            Retry { steps } => self.check_steps(steps),
+            Timeout { steps, .. } => self.check_steps(steps),
+            Parallel { cases, .. } => {
+                let mut errors = Vec::new();
+                cases.iter().for_each(|case| errors.extend(self.check_steps(case)));
+                errors
+            }
+            Call {
+                workflow: workflow_ident,
+                cases: _,
```
rkuhn, line 795: cases need to be checked as well

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,116 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+                ..
+            } => {
+                if let Some(workflow) = self.query.workflows.get(&*workflow_ident) {
+                    self.check_workflow(workflow)
```
rkuhn, line 799: I wouldn’t do this here, since we have a reliable top-level enumeration of all workflows. This means the tracker isn’t needed.

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,116 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+                let mut errors = Vec::new();
+                cases.iter().for_each(|steps| errors.extend(self.check_steps(steps)));
+                errors
+            }
+        }
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct WorkflowTracker<'a>(pub(crate) RefCell<BTreeSet<&'a Workflow<'a>>>);
```
rkuhn, line 823: Probably no longer needed, but still: why `pub(crate)`? Such internal things should always stay as local as possible, especially with a RefCell inside.

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,116 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+}
+
+pub(crate) enum QueryWorkflowAnalysisErrorType {
+    UndeclaredEvent,
+    UndeclaredWorkflowCall,
+}
+
+pub(crate) struct QueryWorkflowAnalysis<'a> {
+    pub(crate) query: &'a Query<'a>,
+    pub(crate) workflow_tracker: WorkflowTracker<'a>,
```
rkuhn, line 737: I think neither of these fields are needed: the (recursive) analysis functions can just get the outer `Query` and the syntax element in question as arguments. When analysing the flow inside a workflow we’ll probably need to have some more context, though, so it might be appropriate to keep this struct with the `query` field in any case.

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: Ah yes.

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: I would also imagine the type checker information to be stored in this struct too because it will be used inside while checking binders

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,100 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+    fn check_workflow(&'a self, workflow: &'a Workflow) -> Vec<QueryWorkflowAnalysisError> {
+        self.check_steps(&workflow.steps)
+    }
+
+    fn check_steps(&'a self, steps: &'a NonEmptyVec<WorkflowStep<'a>>) -> Vec<QueryWorkflowAnalysisError> {
+        steps.iter().flat_map(|step| self.check_step(step)).collect()
+    }
+
+    fn check_step(&'a self, step: &'a WorkflowStep<'a>) -> Vec<QueryWorkflowAnalysisError> {
+        use super::workflow::WorkflowStep::*;
```
jmg-duarte, line 761: Nitpick: Not a fan of using all the enum members. In this case is not particularly problematic but take a look into this https://youtu.be/8j_FbjiowvE?si=DOchI2_0vieFkEjg&t=103 (not the full video)

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,100 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+        }
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct WorkflowTracker<'a>(RefCell<BTreeSet<&'a Workflow<'a>>>);
+
+impl<'a> WorkflowTracker<'a> {
+    fn contains(&'a self, workflow: &'a Workflow) -> bool {
+        self.0.borrow_mut().contains(workflow)
```
jmg-duarte, line 811: Do you need `borrow_mut` to perform `contains`?

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,100 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+                errors.extend(self.check_steps(with));
+                errors
+            }
+            Choice { cases } => cases.iter().flat_map(|steps| self.check_steps(steps)).collect(),
+        }
+    }
+}
+
+#[derive(Default)]
+pub(crate) struct WorkflowTracker<'a>(RefCell<BTreeSet<&'a Workflow<'a>>>);
```
jmg-duarte, line 807: I'm curious as to why a `RefCell`, if this is an analysis pass, this should simply perform the analysis, "lock it" and move on. 

What am I missing?

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,100 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+            } => {
+                let mut errors = Vec::new();
+                if let None = self.query.workflows.get(&*workflow_ident) {
+                    errors.push(QueryWorkflowAnalysisError {
+                        error_type: QueryWorkflowAnalysisErrorType::UndeclaredWorkflowCall,
+                        span: workflow_ident.clone(),
+                    });
+                }
+
+                errors.extend(cases.iter().flat_map(|(_, steps)| self.check_steps(steps)));
```
jmg-duarte, line 791: Since this always runs, you can do:
```rust
let mut errors = cases.iter().flat_map(|(_, steps)| self.check_steps(steps)).collect::<Vec<_>>();
```

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,100 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+            Retry { steps } => self.check_steps(steps),
+            Timeout { steps, .. } => self.check_steps(steps),
+            Parallel { cases, .. } => cases.iter().flat_map(|case| self.check_steps(case)).collect(),
+            Call {
+                workflow: workflow_ident,
+                cases,
+                ..
+            } => {
+                let mut errors = Vec::new();
+                if let None = self.query.workflows.get(&*workflow_ident) {
```
jmg-duarte, line 784:
```suggestion
                if self.query.workflows.get(&*workflow_ident).is_none() {
```

IIRC clippy even complains about this

rust/actyx/ax-aql/src/language/parser.rs
```diff
@@ -721,6 +722,100 @@ pub(crate) fn query_from_str(s: &str) -> Result<Query<'_>> {
...
+                }
+
+                errors.extend(cases.iter().flat_map(|(_, steps)| self.check_steps(steps)));
+
+                errors
+            }
+            Compensate { body, with } => {
+                let mut errors = Vec::new();
+                errors.extend(self.check_steps(body));
+                errors.extend(self.check_steps(with));
```
jmg-duarte, line 798: Nitpick/Note: By "hand-inlining" the call to `check_steps` and using `.chain` you can reduce the 3 vec allocations to 1.

jmg-duarte (commented):
I was faced with similar vec vs iter decisions while writing the code to dive down the workflow getting the steps. Curious to learn your take on that.

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: Never mind this. It was used to remember which workflow had been analyzed. But now it is not used again.
The reason for the RefCell itself is so that a non-mutable reference can be passed recursively to eliminate headache from the possibility of multiple `mut` references

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: nvm this struct

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: wat??

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: For some reason, I feel that workflow_identity error should go first before the `cases` error. Not that I am sure why this ordering matter at all

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: Thanks. I should be running cargo clippy

rust/actyx/ax-aql/src/language/parser.rs
Kelerchian: What do you mean hand inlining?

Kelerchian:
> I was faced with similar vec vs iter decisions while writing the code to dive down the workflow getting the steps. Curious to learn your take on that.

My first thought was: if these methods can return an iterator rather than vec, then at the end a chain and collect is executed, then it might result in a better code. But I do not know how to do that nor I can find any references on the internet on how to do that

rust/actyx/ax-aql/src/language/parser.rs
jmg-duarte: Copy-pasting the code here 😅

Kelerchian: Do you mean like:
```rust
self
  .check_steps(body)
  .iter()
  .chain(self.check_steps(with).iter())
  .collect()
```

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -184,55 +183,365 @@ impl<'a> QueryTypeCheck<'a> {
...
+                    (Err(a), Ok(_)) => Err(a),
+                    (Err(a), Err(b)) => Err([a, b]
+                        .into_iter()
+                        .map(|err_msg| format!("{}", err_msg))
+                        .collect::<Vec<_>>()
+                        .join("\n")),
+                }
+            }
+            Type::Intersection(_) => {
+                // NOTE: impossible to drill down on intersection without intersection collapsing mechanism
```
jmg-duarte, line 254: Is this a note or a TODO?

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -184,55 +183,365 @@ impl<'a> QueryTypeCheck<'a> {
...
-                crate::TypeAtom::Tuple(x) => match first {
-                    Index::Number(num_index) => {
-                        if (*num_index as usize) < x.len() {
-                            Ok(())
-                        } else {
-                            Err(format!("{:?} cannot be accessed by {}", x, num_index))
-                        }
-                    }
-                    _ => Err(format!("{:?} cannot be accessed by {:?}", x, first)),
+            Type::Atom(type_atom) => match type_atom {
```
jmg-duarte, line 195: Nitpick: I'd make this `Type::Atom(TypeAtom::Universal)` (for example) so you could avoid a whole level of nesting and get "extra context" when reading the match arm

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -184,55 +183,365 @@ impl<'a> QueryTypeCheck<'a> {
...
-        let (first, rest) = index.split_at(1);
-        if first.is_empty() {
-            return Ok(());
+fn drill_down_type_by_index(cur_type: &Type, Ind { head: _, tail, .. }: &Ind) -> Result<Type, ExtraMessage> {
+    fn recurse(cur_type: &Type, index: &[Index]) -> Result<Type, ExtraMessage> {
+        if index.len() < 1 {
+            return Ok(cur_type.to_owned());
         }
-
+        let (first, rest) = index.split_at(1);
```
jmg-duarte, line 191: Why not
```rust
let first = &index[0];
let rest = &index[1..];
```

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -184,55 +183,365 @@ impl<'a> QueryTypeCheck<'a> {
...
+            | SimpleExpr::TimeLiteral(_)
+            | SimpleExpr::Tags(_)
+            | SimpleExpr::App(_)
+            | SimpleExpr::BinOp(_) => return Err(format!("{:?} cannot be accessed by {}", cur_type, first_expr)),
+            SimpleExpr::Variable(_)
+            | SimpleExpr::Indexing(_)
+            | SimpleExpr::Interpolation(_)
+            | SimpleExpr::AggrOp(_)
+            | SimpleExpr::FuncCall(_)
+            | SimpleExpr::SubQuery(_) => return Ok(Type::Atom(TypeAtom::Universal)),
```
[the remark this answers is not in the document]
Kelerchian: I'm avoiding catch-all because there's a chance SimpleExpr will be expanded

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -0,0 +1,575 @@
+use super::workflow::{Participant, Workflow, WorkflowStep};
+use crate::{Ident, Ind, Index, Label, NonEmptyVec, Num, Query, SimpleExpr, Type, TypeAtom};
+use std::{collections::BTreeSet, ops::Deref, sync::Arc};
+pub(crate) struct QueryTypeCheckError<'a> {
+    pub(crate) span: pest::Span<'a>,
+    pub(crate) error_type: QueryTypeCheckErrorType,
+}
+
+type ExtraMessage = String;
```
rkuhn, line 9: This is usually not done in Rust: `type` aliases are used to write down complex types with less verbosity, otherwise newtypes like `struct ExtraMessage(String)` are used to create a separate type with different semantics.

Kelerchian: With `struct` how can I make it so that `ExtraMessage` can be operated as if it is a string?

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -184,55 +183,365 @@ impl<'a> QueryTypeCheck<'a> {
...
-                        Err(format!("{:?} cannot be accessed by {:?}", x, first))
+                        Err(format!("{:?} cannot be accessed by {:?}", type_atom, first))
                     }
                 }
-                _ => Err(format!("{:?} cannot be accessed by {:?}", x, first)),
+                _ => Err(format!("{:?} cannot be accessed by {:?}", type_atom, first)),
             },
             Type::Dict(ty) => match first {
+                Index::Expr(expr) => recurse_with_calculated_expression(cur_type, expr, rest),
                 Index::String(_) => recurse(ty, rest),
```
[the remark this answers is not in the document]
Kelerchian: ah you're right.

rust/actyx/ax-aql/src/language/type_check.rs
Kelerchian: I want to state that this is a string, but also an "ExtraMessage" for `InvalidIndex` error.
The reason is, `DrillDownResult` cannot be `Result<Type, QueryTypeCheckErrorType::InvalidIndex>` and `Result<Type, String>` isn't clear enough on what the string are

rkuhn: In that case I’d just use `Result<Type, QueryTypeCheckErrorType>` as the return type.

rust/actyx/ax-aql/src/language/type_check.rs
```diff
@@ -0,0 +1,575 @@
...
+                        Participant::Role(ident) => ident,
+                        Participant::Unique(ident) => ident,
+                    }
+                })
+                .collect(),
+        }
+    }
+}
+
+pub(crate) struct QueryTypeCheck<'a> {
```
rkuhn, line 40: Overall, type checking is a function, not an object, so I’d start from functions, not methods. If the argument lists become too long and repetitive, then some arguments can be passed within a single struct.

rkuhn (commented):
In general the control flow structure will need to be geared towards traversing the statements of a workflow in the same order as during evaluation, including a context that is enriched along the way. This enables checking that `eventLabel @ whatever` refers to a valid `whatever` — which may either be a workflow argument or a binding.

rust/actyx/ax-aql/src/language/type_check/drill_down_type.rs
```diff
@@ -0,0 +1,443 @@
...
+                if !errors.is_empty() {
+                    let error_messages = errors
+                        .into_iter()
+                        .map(|err_msg| format!("- {}", err_msg))
+                        .collect::<Vec<_>>()
+                        .join("\n");
+                    return Err(error_messages.trim().into());
+                }
+
+                // NOTE: Using Option to emulate bottom type; consider introducing bottom type
```
jmg-duarte, line 179: Is `NO_VALUE` the bottom type? Maybe this could be addressed with it?

rust/actyx/ax-aql/src/language/type_check/drill_down_type.rs
```diff
@@ -0,0 +1,443 @@
...
+        );
+
+        assert!(drill_down_type_by_index(
+            &Type::Array(Arc::new(Type::Atom(TypeAtom::Number(None)))),
+            &Ind {
+                head: SimpleExpr::Variable(Var("_".to_string())).into(),
+                tail: vec![Index::String(String::from("some_string"))].try_into().unwrap(),
+            },
+        )
+        .is_err());
```
jmg-duarte, line 258: I'd add the second parameter to `assert!` with a short explanation of what failed

rust/actyx/ax-aql/src/language/type_check/drill_down_type.rs
```diff
@@ -0,0 +1,443 @@
...
+                    .try_into()
+                    .unwrap(),
+                },
+            ),
+            Ok(Type::Atom(TypeAtom::Null))
+        );
+    }
+
+    #[test]
+    fn tuple() {
```
jmg-duarte, line 328: Nit: I'd separate all the asserts in this test into multiple tests. AFAIK you also get to take more advantage of parallelism that way

rust/actyx/ax-aql/src/language/type_check/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+                        Participant::Role(ident) => ident,
+                        Participant::Unique(ident) => ident,
+                    }
+                })
+                .collect(),
+        }
+    }
+}
+
+pub(crate) struct QueryTypeCheck<'a> {
```
jmg-duarte, line 42: I don't see how this structure helps the overall goal. Why not use just the `Query<'a>`

rust/actyx/ax-aql/src/language/type_check/mod.rs
jmg-duarte, line 1: I think this file is lacking some docs, they should help you pick the file back up in the future too as well as any potential contributors.

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+}
+
+impl ToString for CollapseError {
+    fn to_string(&self) -> String {
+        self.0.clone().join(". ")
+    }
+}
+
+impl Type {
+    /// Build union out of DoubleEndedIterator of types
```
jmg-duarte, line 42: Reading this and looking at the method I am left just as wise about it. Is it just creating a `Union` of all types on the iterator?

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the iterator does not yield any value!
+    pub(crate) fn union(iter: impl DoubleEndedIterator<Item = Type>) -> Type {
+        let rebuilt = iter.into_iter().rev().fold(None, |acc, next| match acc {
+            None => Some(next.clone()),
+            Some(prev) => Some(Type::Union(Arc::new((prev, next.clone())))),
```
jmg-duarte, line 50: Don't we have ownership of the types? Why clone them?

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+        });
+
+        rebuilt.expect("impossible for union to reduce its subtypes to zero")
+    }
+
+    /// Attempt to apply collapse and sorts to intersections, unions, records
+    pub(crate) fn collapse(self) -> Result<Type, CollapseError> {
```
jmg-duarte, line 57: I think this is usually called unification, see: https://go.dev/blog/type-inference#solving-type-equations

Furthermore, the documentation doesn't really explain/define what collapsing is, just that the method collapses things.

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+    fn flatten_union_collapsing((a, b): &(Type, Type)) -> Result<BTreeSet<Type>, CollapseError> {
+        let mut under_types_vec = vec![a.clone(), b.clone()];
+        let mut collapse_errors = vec![];
+
+        loop {
+            let mut non_unions = vec![];
+            let mut unions = vec![];
+
+            under_types_vec
+                .split_off(0)
```
jmg-duarte, line 236: Isn't this a no-op?

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+                }
+
+                match (&a, &b) {
+                    // all intersections should have been collapsed
+                    (Type::Intersection(_), _) => Err(CollapseError(vec![
+                        "intersection could not exist after a collapse".into(),
+                    ])),
+                    (_, Type::Intersection(_)) => Err(CollapseError(vec![
+                        "intersection could not exist after a collapse".into(),
+                    ])),
```
jmg-duarte, line 79:
```suggestion
                    (Type::Intersection(_), _) | (_, Type::Intersection(_)) => Err(CollapseError(vec![
                        "intersection could not exist after a collapse".into(),
                    ])),
```

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+                    // intersection of records
+                    (Type::Atom(TypeAtom::Record(a)), Type::Atom(TypeAtom::Record(b))) => {
```
jmg-duarte, line 81: This is using the old model. I moved the record away from `TypeAtom`. Not sure about the repercussions for this instance in particular

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+                                    }
+                                }
+                                Err(err) => errors.push(err),
+                            });
+
+                        if !errors.is_empty() {
+                            return Err(CollapseError::join(errors));
+                        }
+
+                        if !conflicts.is_empty() {
```
jmg-duarte, line 111: Shouldn't the errors here be checked alongside `errors`? As a user I prefer receiving all possible errors at once instead of "unlocking them"

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+            (Type::Union(a), Type::Union(b)) => {
+                let a = Type::flatten_union(a.as_ref());
+                let b = Type::flatten_union(b.as_ref());
+
+                a.intersection(&b).cloned().collect::<BTreeSet<_>>() == b
+            }
+            (Type::Union(union), non_union) => Type::flatten_union(union.as_ref()).contains(non_union),
+            (non_union, Type::Union(union)) => Type::flatten_union(union.as_ref()).contains(non_union),
+            (Type::Array(a), Type::Array(b)) => a.is_supertype_of(b.as_ref()),
+            (Type::Dict(a), Type::Dict(b)) => a.as_ref().is_supertype_of(b.as_ref()),
```
jmg-duarte, line 194: Is the first `as_ref` necessary?

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -8,17 +8,258 @@ use crate::{
...
+                    a.intersection(&b).cloned().collect::<BTreeSet<_>>() == b
+                }
+                (TypeAtom::Universal, _) => true,
+                _ => false,
+            },
+            (Type::Union(a), Type::Union(b)) => {
+                let a = Type::flatten_union(a.as_ref());
+                let b = Type::flatten_union(b.as_ref());
+
+                a.intersection(&b).cloned().collect::<BTreeSet<_>>() == b
```
jmg-duarte, line 189: I'm not convinced about this, here's a counter example:


* Let `a` be `NUMBER | BOOLEAN`
* Let `b` be `NUMBER(10) | NULL`
* `a & b => NUMBER(10)`
* `(a & b) != b`

Let me know if I'm missing something

rust/actyx/ax-aql/src/language/types.rs
Kelerchian: nice catch! this piece of code was copied from another place. thanks

rust/actyx/ax-aql/src/language/types.rs
Kelerchian: yep. you said it better than the comment tho. changing the comment

rust/actyx/ax-aql/src/language/types.rs
Kelerchian: when there is a collapse error, conflict detection is no longer reliable and is skipped. this is why Err from errors comes first before conflicts

rust/actyx/ax-aql/src/language/type_check/drill_down_type.rs
Kelerchian: no, no_value is not the subtype of NULL or NUMBER for example

rust/actyx/ax-aql/src/language/type_check/mod.rs
Kelerchian: this should have been removed. nice catch

rust/actyx/ax-aql/src/language/types.rs
Kelerchian: the goal is to take all members of the vec and leave it empty.
i should replace this with `std::mem::take(&mut under_types_vec)` tho

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -25,6 +33,319 @@ pub enum Type {
...
+            Type::Union(union) => {
+                let (a, b) = union.as_ref();
+                Spread::Many(vec![a.clone(), b.clone()])
+            }
+            x => Spread::One(x),
+        });
+
+        spread.into_iter().collect()
+    }
+}
```
jmg-duarte, line 347: I feel there must be a better way of writing this API. The double clone that ends up happening in line 337 seems wrong.

Maybe you could use a `Collapse` trait and implement it for the various `BTreeSet`, delaying the cloning until its really necessary for example. Furthermore, as we discussed, `collapse` probably doesn't need to consume the type, you're just checking a bunch of things and returning a new type.

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -25,6 +33,319 @@ pub enum Type {
     Record(NonEmptyVec<(Label, Type)>),
 }
 
+enum Spread<T> {
+    Many(Vec<T>),
+    One(T),
+}
+
+fn spread<T>(vec: Vec<T>, spreader: impl Fn(&T) -> Spread<T>) -> Vec<T> {
```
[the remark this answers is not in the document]
Kelerchian: will do. and it will be moved down because it doesn't make sense being between struct Type and impl Type

rust/actyx/ax-aql/src/language/types.rs
```diff
@@ -25,6 +33,319 @@ pub enum Type {
...
+                // subtypes
+                (TypeAtom::Bool(Some(_)), TypeAtom::Bool(None)) => Hierarchy::Subtype,
+                (TypeAtom::Number(Some(_)), TypeAtom::Number(None)) => Hierarchy::Subtype,
+                (TypeAtom::String(Some(_)), TypeAtom::String(None)) => Hierarchy::Subtype,
+                _ => Hierarchy::Disjointed,
+            },
+            (Type::Record(a), Type::Record(b)) => {
+                let a = a.iter().collect::<BTreeSet<_>>();
+                let b = b.iter().collect::<BTreeSet<_>>();
+                let intersection = BTreeSet::intersection(&a, &b).cloned().collect::<BTreeSet<_>>();
```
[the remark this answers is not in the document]
rkuhn: I haven’t yet read the code, but @jmg-duarte your comment above is incorrect: a subtype can be used in place of its supertype (Liskov substitution principle), which means that it must have at least the properties of the supertype. A is a subtype of B, in both cases.

rust/actyx/ax-aql/src/language/types.rs
Kelerchian: while collapse doesn't consume, it still produces a new type.
this in turn forces this function to return BTreeSet<Type> and not BTreeSet<&Type>
because we cannot return references to objects created within the function (produced by collapse)

rust/actyx/ax-aql/src/language/types.rs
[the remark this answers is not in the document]
Kelerchian: really nice catch. thanks  for this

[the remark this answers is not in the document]
Kelerchian: https://www.cs.cornell.edu/courses/cs6110/2018sp/lectures/lec23.pdf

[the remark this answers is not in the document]
jmg-duarte: > I haven’t yet read the code, but @jmg-duarte your comment above is incorrect: a subtype can be used in place of its supertype (Liskov substitution principle), which means that it must have at least the properties of the supertype. A is a subtype of B, in both cases.

True, I applied the logic for refinements here where NUMBER is supertype of 10. 

But I think the nesting issue still applies?

rust/actyx/ax-aql/src/language/types.rs
jmg-duarte: Yeah that makes sense but my point was also about how (if possible) to make a slightly better API by just having a single function instead of these two.

rust/actyx/ax-aql/src/language/types.rs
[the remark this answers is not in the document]
Kelerchian: The nesting issue still applies I think.

PR #685, opened 2024-03-15
Add concurrency support for CI

jmg-duarte:
This should help reduce netsim errors and CI load when pushing multiple consecutive updates to the same PR

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -8,6 +8,11 @@ on:
     branches:
       - "master"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Only cancel for pull requests
+  cancel-in-progress: ${{ github.event_name == 'pull_request'}}
+
 env:
   RUST_BACKTRACE: 1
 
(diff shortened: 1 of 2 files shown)


.github/workflows/validation.yml
```diff
@@ -8,6 +8,12 @@ on:
     branches:
       - "master"
 
+concurrency:
+  # If there's a PR number, it will result in cancelling the previous build for that PR
+  # this is needed because the PR event triggers on "meta" PR things, like comments but not pushes
+  group: ${{ github.workflow }}-${{ github.ref_name }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
```
rkuhn, line 15: There’s probably also a “number of concurrent jobs” setting that defaults to 1, right? If so, please make it explicit.

.github/workflows/validation.yml
```diff
@@ -50,20 +56,33 @@ jobs:
           make validate-os-android assert-clean
       - run: make clean
 
-  validate:
+  validate-os:
```
rkuhn, line 59: at this juncture we might was well rename it to `validate-core`, WDYT?

rkuhn (approved):
Looking good!

.github/workflows/validation.yml
jmg-duarte: Sure, I just kept os because it's the command in the Makefile — `make validate-os`

.github/workflows/validation.yml
jmg-duarte: You'd think so, but there's no setting like that
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

PR #686, opened 2024-03-20
Handle leftover issues from #683

jmg-duarte:
Handles:
> - if these errors are to be read by humans, they would be vastly more useful if they carried the context (i.e. the path inside the event payload object)
> - since these errors will usually not be read by humans and they are expensive to construct (string formatting), it would be really great if the error carried the information for printing but only did it in its Display impl

— _https://github.com/Actyx/Actyx/pull/683#discussion_r1528236382_

### DIFF
--- a/rust/actyx/Cargo.lock
+++ b/rust/actyx/Cargo.lock
@@ -1119,9 +1119,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cbor-data"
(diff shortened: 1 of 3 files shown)


rkuhn (commented):
Yes, this is a good first step! Now make it so that constructing a TypeError does not need allocation (i.e. no `.clone()`, `.to_string()` etc.) by only referencing the data needed for printing.

`with_path` is the only place where allocation really is required because of the dynamic nature. Here, it can be minimised by using a SmallVec and storing `&'a Index` (or a new enum if required).

Issue #687, opened 2024-04-11
[Bug]: Names containing underscore is not accepted as a node hostname


Kelerchian:
### Product

Node Manager

### Operating System

None

### Current behavior

(as described in the title)

### Expected behavior

Names containing underscore is valid

PR #689, opened 2024-04-15
fix(node-manager): #687 names containing underscore is not accepted a…

Kelerchian:
…s node hostname


rkuhn:
Please move the changelog entry away from the first line of the commit comment, because if we merge this then the changelog will be duplicated in the merge commit.

Kelerchian:
> Please move the changelog entry away from the first line of the commit comment, because if we merge this then the changelog will be duplicated in the merge commit.

I missed this, my bad. How do I fix this? Does a revert work?

rkuhn:
No, now a force push is needed — will do it later.

